PR #46, opened 2015-02-15
[#2698] Remove twisted dependency.

adiroiban:
Problem
=======

compat has a stupid dependency on Twisted filepath and we need to remove it

Why we got this
============

In the first version of compat, it tried to reuse the Twisted Filepath code... but later found out that Filepath is to simple and not cross platform.

Changes
======

use path directly ,not via Twisted filepath

How to test
========

reviewer: @alibotean 

Thanks!


adiroiban:
needs-review

alibotean:
Tests are failing on two machines:

* AIX http://build.chevah.com/builders/compat-aix-71/builds/28
* Linux http://build.chevah.com/builders/compat-linux-x86/builds/6/steps/deps/logs/stdio

Codewise the changes look ok.

alibotean:
approved-at latest

PR #48, opened 2015-03-11
[#2704] Update support for HPUX

adiroiban:
Problem
======

hpux is no longer a recognized OS in compat.

Why we got this
=============

We no longer have a buildslave for HPUX and latest versions were not tested on hpux.

Changes
=======

Update pavement.py to use valid version numbers as hpux was tested on HP CLOE external servers.

update OS administration for  HPUX.
Disable tests and create dedicated tickets for each issue.

How to test
=========

reviewers: @alibotean 

we now have our own hpux buildslave and once this is merge we  can enable the tests for PQM


adiroiban:
needs-review

alibotean:
changes-approved

PR #49, opened 2015-04-17
[#2815] Fix handling of Unicode and with spaces arguments on Windows

adiroiban:
Problem
=======

We can not receive unicode arguments in Windows from command line.

From our batch file we can not receive arguments with spaces.

Changes
=======

Use Unicode windows API to parse arguments.

Python 2.7 API is broken  and I can not use it to automate this tests as it fails to execute external command with unicode arguments 

To fix
====

Passing arguments with spaces from batch file is not yet fixed.

How to test
=========

reviewers: @alibotean 

```
paver build
./build-windows-x86/lib/python -m chevah.compat.tests.manual.print_argv ARG or NO ARG
```

from a plain cmd.exe
```
cd build-window-x86
..\chevah\compat\tests\manual\batch_argv.bat ARG or NO ARG
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -15,6 +15,9 @@ and Windows, see IOSUsers.
 
 Support for staring Unix daemon and installing and starting Windows services.
 
+Importing this module on Windows will populate the sys.argv with Unicode
+values.
+
 Unified interface for working with operating system file systems provides:
 
 * Unicode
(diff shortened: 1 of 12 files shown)


adiroiban:
I wrote an automated tests but I am not sure it is valid since it uses Python to launch python itself and it requires encoding the path before calling process open

adiroiban:
Thanks for the feedback.

The case for no argument will not help.

I placed a single ^ and looks like it is working.

needs-review

alibotean:
> I placed a single ^ and looks like it is working.

Yuhu :smile:

changes-approved

PR #50, opened 2015-04-26
[#2831] Update code to support python3.

adiroiban:
Problem
=======

We want to migrate to python3

Changes
=======

Update code and add test_py3 step

I have used to automatic migration as described here

https://trac.chevah.com:10443/wiki/Development/Python#python3-porting

some imports are not really needed ... like division or absolute or print function as the code does not use those features... but I left them for future changes.. and since it was more work to manually hunt for the places which need them.

some python2 API like grp module and PAM on OSX does not work with bytes.

On python2 bytes is just an alias for 'str'. This mean that you can perform text like operation on bytes.

On python3 bytes are a separate type with restricted operations.

With future `str` type the py2 str is conveted into a type similar to py3 str ... which is always unicode and when encoded will return bytes 

How to test
========

reviewers: @alibotean 

Thanks!


adiroiban:
needs-review

chevah/compat/administration.py
```diff
@@ -775,7 +782,9 @@ def deleteUser(self, user):
         import win32net
         try:
             win32net.NetUserDel(user.pdc, user.name)
-        except win32net.error, (number, context, message):
+        except win32net.error as xxx_todo_changeme:
```
alibotean: xxx_todo_changeme? :smile:

chevah/compat/nt_users.py
```diff
@@ -180,7 +184,8 @@ def userExists(self, username):
         try:
             win32security.LookupAccountName('', username)
             return True
-        except win32security.error, (number, name, message):
+        except win32security.error as xxx_todo_changeme1:
```
alibotean: same as above

pylintrc
```diff
@@ -0,0 +1,320 @@
...
+[MESSAGES CONTROL]
+
+# Valid levels: HIGH, INFERENCE, INFERENCE_FAILURE, UNDEFINED
+confidence=
+
+# Enable the message, report, category or checker with the given id(s). You can
+# either give multiple identifier separated by comma (,) or put this option
+# multiple time. See also the "--disable" option for examples.
+#disable=all
+enalbe=python3
```
alibotean: enalbe instead of enable

alibotean:
Just some minor comments, please take a look.

changes-approved

PR #51, opened 2015-05-22
[#2845] Fix userExists on Unix to read /etc/passwd as normal user.

adiroiban:
Problem
=======

/etc/passwd can be read by anyone on Unix ... but compat read it as root

Changes
=======

Read file as service account.

How to test
=========

reviewers: @alibotean 

thanks!

### DIFF
--- a/chevah/compat/unix_users.py
+++ b/chevah/compat/unix_users.py
@@ -129,12 +129,11 @@ class UnixUsers(CompatUsers):
             return False
 
         username = username.encode('utf-8')
-        with self._executeAsAdministrator():
-            try:
-                pwd.getpwnam(username)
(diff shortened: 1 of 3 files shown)


alibotean:
Looks ok to me.

changes-approved

I see that two tests are failing on sles-11 and they do not look like transient errors:

http://build.chevah.com/builders/compat-sles-11/builds/227/steps/test/logs/stdio

adiroiban:
yes... the sles build is a bit messed up. but not due to this branch. I am looking to fix the sles slave and will merge.

thanks!

adiroiban:
problem is described here https://trac.chevah.com:10443/ticket/2857

PR #52, opened 2015-10-09
[#3044] Add node_id and comparison to IFileAttributes.

adiroiban:
Scope
=====

This branch updated the IFileAttributes to include node_id , owner and group members.
It also updates the FileAttributes implementation to support comparisons.

Changes
=======

The node_id is already available under Unix/Linux and was extend for Windows.

uid and gid are ok for local filesystem nodes but for example remote FTP files don't have uid/gid but just owne / group names.

0.31.0 was already publish as it is required by chevah.server

How to test
========

reviewers: @alibotean 

You can create a hard link on Windows and then when getting status for both paths you should see the same id

### DIFF
--- a/chevah/compat/interfaces.py
+++ b/chevah/compat/interfaces.py
@@ -578,9 +578,14 @@ class IFileAttributes(Interface):
     is_file = Attribute('True if member is a file.')
     is_folder = Attribute('True if member is a folder.')
     is_link = Attribute('True if member is a symbolic link.')
-    modified = Attribute('Timestamp at which content modified.')
+    modified = Attribute('Timestamp at which content was last modified.')
 
     mode = Attribute('Protection bits. Unix specific.')
     hardlinks = Attribute('Number of hard links.')
+
     uid = Attribute('User ID or owner as integer.')
-    gid = Attribute('Group ID of owner as integer.')
+    gid = Attribute('Group ID as integer.')
+
+    node_id = Attribute('ID inside the filesystem.')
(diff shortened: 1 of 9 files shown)


adiroiban:
needs-review

alibotean:
Looks like all the attributes are the same, not just the node id.

Python:

```
>>> from chevah.compat.nt_filesystem import *
>>> fs = NTFilesystem()
>>> k=['c', 'temp', 'a.txt']
>>> print fs.getStatus(k)
nt.stat_result(st_mode=33206, st_ino=36560001433286537107596218741L, st_dev=0, st_nlink=0, st_uid=0, st_gid=0, st_size=8L, st_atime=1444664963L, st_mtime=1444664964L, st_ctime=1444664963L)
>>> kk=['c','temp','b.lnk']
>>> print fs.getStatus(kk)
nt.stat_result(st_mode=33206, st_ino=36560001433286537107596218741L, st_dev=0, st_nlink=0, st_uid=0, st_gid=0, st_size=8L, st_atime=1444664963L, st_mtime=1444664964L, st_ctime=1444664963L)
>>>
```

Command prompt:

```
c:\temp>mklink /H b.lnk a.txt
Hardlink created for b.lnk <<===>> a.txt

c:\temp>ls
a.txt  b.lnk
```

alibotean:
changes-approved

adiroiban:
> Looks like all the attributes are the same, not just the node id.

This is the expected behaviou. Hardlinks are just multiple names for the same data.

Thanks for the review.

PR #53, opened 2015-10-18
[#3057] Add dedicated PAM method to system_users.

adiroiban:
Scope
=====

This adds a public method to system_users for auhentication accounts using PAM.

Changes
======

Export the PAM method. Extend it to use a custom service name.

Remove the LDAP tests as they were there to test PAM. Use dedicated PAM module for testing chevah.compat.

/etc/password is readable by everyone so we don't need to read it as root.

PAM can also be used without root

I had to use root when calling PAM as otherwise it was failing when calling with sudo ?!
I have no idea why.... locks like some logging issue . ..

Since PAM will authenticate OS accounts the process will need anyway to check that the account can be impersonated. 

How to test
=========

reviewers: @alibotean @dumol 

check that changes make sense

not all test pass as the new PAM module is not yet distributed on all system.

There is a dedicated ticket https://trac.chevah.com:10443/ticket/3058

@dumol you can use this branch to check the result of ticket 3058

Thanks!


adiroiban:
needs-reivew

dumol:
This doesn't show up in my Trac, probably because of the misspelling above…

**needs-review**

dumol:
Some minor comments above in regard to the comments.

**changes-approved**

alibotean:
changes-approved

PR #55, opened 2015-11-23
[#3126] Add dedicated tests for passwd / shadow and pam

adiroiban:
Scope
======

This updated the test for Unix password methods.

The previous one were high level and they did not provide fine grained control for passwd/shadow/pam methods

Changes
=======

Refactor code and add dedicated tests for each unix auth method.

With this I found an error in Solaris PAM. There is a dedicated ticket

I have also found the special  behaviour of AIX with /etc/security/passwd which works like a shadow file.

How to test
========

reviewers: @alibotean 

check that changes make sense

thanks!

### DIFF
--- a/chevah/compat/tests/elevated/test_system_users.py
+++ b/chevah/compat/tests/elevated/test_system_users.py
@@ -245,7 +245,7 @@ class TestSystemUsers(SystemUsersTestCase):
             self.assertIsNotNone(token)
 
     @conditionals.onOSFamily('posix')
-    def test_authenticateWithUsernameAndPassword_passwd_valid(self):
+    def test_checkPasswdFile_valid(self):
         """
         On most OS system password is not stored in the passwd file so even
         if we pass the correct pass it will return None to inform that
@@ -264,7 +264,7 @@ class TestSystemUsers(SystemUsersTestCase):
             self.assertIsNone(result)
 
     @conditionals.onOSFamily('posix')
-    def test_authenticateWithUsernameAndPassword_passwd_invalid(self):
+    def test_checkPasswdFile_invalid(self):
         """
         When an invalid password is provided, on most OS system password is not
         stored in the passwd file it return None to inform that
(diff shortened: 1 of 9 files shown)


adiroiban:
needs-review

chevah/compat/tests/elevated/test_system_users.py
```diff
@@ -244,6 +244,84 @@ def test_authenticateWithUsernameAndPassword_good(self):
...
+        if self.os_name in ['aix', 'hpux']:
+            # AIX and HPUX has no shadow
+            raise self.skipTest()
+
+        result = system_users._checkShadowFile(
+            username=TEST_ACCOUNT_USERNAME,
+            password=TEST_ACCOUNT_PASSWORD,
+            )
+
+        if self.os_name in ['aix', 'hpux', 'osx']:
```
alibotean: no point in having `aix`/`hpux` since the test is skipped

chevah/compat/tests/elevated/test_system_users.py
```diff
@@ -244,6 +244,84 @@ def test_authenticateWithUsernameAndPassword_good(self):
         else:
             self.assertIsNotNone(token)
 
+    @conditionals.onOSFamily('posix')
+    def test_authenticateWithUsernameAndPassword_passwd_valid(self):
```
alibotean: Is there a reason for not changing the test method name to reflect the actual method tested?

alibotean:
Just minor comments.

changes-approved

chevah/compat/tests/elevated/test_system_users.py
adiroiban: thanks... test should not be skipped, but rather check that shadow is not supported

chevah/compat/tests/elevated/test_system_users.py
adiroiban: bad coder :)

fixed. thanks!

PR #56, opened 2015-11-24
[#3132] Auth OS accounts only based on PAM on AIX.

adiroiban:
Scope
=====

This is required for a specific sale to RoyalMail

They have an AIX system in which /etc/password is in conflict  with PAM.

This will disable /etc/passwd auth on AIX.

Change
=======

just disable for AIX

there is already a branch to add a separate PAM auth ... but we need more work to merge it

How to test
=========

reviewers: @alibotean 

thanks

### DIFF
--- a/chevah/compat/unix_users.py
+++ b/chevah/compat/unix_users.py
@@ -217,6 +217,11 @@ class UnixUsers(CompatUsers):
         Return False if user was not found or password is wrong.
         Returns None if password is stored in shadow file.
         '''
+        from chevah.compat import process_capabilities
+        if process_capabilities.os_name == 'aix':
+            # AIX /etc/passwd auth is disabled for this release.
+            return None
+
         username = username.encode('utf-8')
         password = password.encode('utf-8')
         try:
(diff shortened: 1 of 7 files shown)


adiroiban:
just a test failure due to a time delta.

needs-review

chevah/compat/unix_users.py
```diff
@@ -217,6 +217,11 @@ def _checkPasswdFile(self, username, password):
         Return False if user was not found or password is wrong.
         Returns None if password is stored in shadow file.
         '''
+        from chevah.compat import process_capabilities
+        if process_capabilities.os_name == 'aix':
+            # AIX /etc/passwd auth is disabled for this release.
```
alibotean: No fixme ref ;)

alibotean:
I suggest we add a fixme ref to know why this is was added and when we plan to remove it.

changes-approved

adiroiban:
Thanks. I plan to revert this ASAP  and land #55 soon after this hit master

PR #57, opened 2016-03-28
[#3370] Add support for FreeBSD.

adiroiban:
Scope
=====

This adds support for freebsd... and also openbsd

Changes
=======

The *BSD support was added without shadow password.

I want to merge it now, as BSD is low prioirty and I don't think that we will add support for it soon... but we don't need to block on that and we can still use it without elevated support

as a drive by I have removed the get_crypted_password which is not used.

I have updated buildbot to run tests on openbsd and freebsd

How to test
=========

reviewers: @brunogola 

the coverate is bad as for now we don't run elevated tests on openbsd and freebsd due to lack of shadow password support

check that changes make sense... once this is master I will enable all buildslaves for compat

### DIFF
--- a/chevah/compat/__init__.py
+++ b/chevah/compat/__init__.py
@@ -11,6 +11,27 @@ from __future__ import absolute_import
 import os
 import sys
 
+# Check stdblib shadow support.
+try:
+    import spwd
+    _HAS_SHADOW_SUPPORT = True
+except ImportError:
+    spwd = None
+    _HAS_SHADOW_SUPPORT = False
+
+# Check if OS is storing users in databases.
+try:
+    import dbm
+    import whichdb
+    db_type = whichdb.whichdb('/etc/pwd')
+    if db_type in ['dbm', 'bsddb185']:
+        _HAS_NDBM_SUPPORT = True
+    else:
+        _HAS_NDBM_SUPPORT = False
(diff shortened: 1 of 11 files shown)


adiroiban:
needs-review

brunogola:
changes-approved

adiroiban:
@brunogola thanks for the review.

please take extra care when doing the review... preferably to do the review first thing in the morning.

I made a very stupid thing here https://github.com/chevah/compat/pull/57/commits/cee10ffc9b494526211b14123a11460cb74332cb#diff-f2b2105cf9a1a085416130965a62337aL319

and it looks it was not detected by the review

brunogola:
you mean the removal of `get_crypted_password` ?

as it was mentioned in the PR description I thought it was ok, but maybe I should have checked if it really was used or not?

adiroiban:
@brunogola I have accidental also removed    ```return _verifyCrypt(password, crypted_password)```

brunogola:
ah, wow, true.

the red color block from github diff is not helping in this case :/

i will pay better attention

adiroiban:
np... it can happen.. I have also not observed the accidental removal and was happy that tests were green.. We should have better tools.  In the follow up PR I am trying to  fix that with a better tests

PR #60, opened 2016-05-17
[#3416] Allow getting node id on Windows even when file is open.

alibotean:
Scope
=====

Our `getStatus` implementation fails with an error for files that are kept open by other processes (or even our process).

We want to be able to get file information details even if the file is not closed yet.

Changes
=======

I've used the same flags as the ones in https://github.com/python/cpython/blob/master/Modules/posixmodule.c#L1499

Done a bit of TDD and added a functional test for this as well.

reviewers: @adiroiban 

Please check that changes make sense. Once all is good I will add a test in the server which will break if we upgrade the compat without fixing the flow. At least this is my understanding that we want to do.

### DIFF
--- a/chevah/compat/nt_filesystem.py
+++ b/chevah/compat/nt_filesystem.py
@@ -32,19 +32,35 @@ from chevah.compat.nt_users import NTDefaultAvatar, NTUsers
 from chevah.compat.posix_filesystem import PosixFilesystemBase
 
 
-# cut-and-pasted from MSDN
+#: https://msdn.microsoft.com/en-us/library/windows/desktop/aa364939.aspx
 # 0 Unknown
 # 1 No Root Directory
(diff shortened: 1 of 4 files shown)


alibotean:
needs-review

adiroiban:
Great stuff.

Please merge.

Server should not break if we update to this version, but in server we might not want to ignore the errors as all imporant errors should already be ignored by compat

Thanks!

changes-approved

alibotean:
> Server should not break if we update to this version, but in server we might not want to ignore the errors as all imporant errors should already be ignored by compat

I was being stupid; server (master) will not break.

My revert PR for the server has failing tests until this is merged and updated in server. Once that is completed the tests will pass -- proving we can get status for files that are opened.

PR #62, opened 2016-07-01
[#3120] Compat: fix Python 2.7.4+ Windows path handling

alibotean:
Scope
=====

Paths that start with `//` are incorrectly handled on Windows.

Something changed starting with Python 2.7.4 release and the `ntpath.normpath` function which we are using is no longer removing the double slashes along with `.` and `..` from the beginning of the path.

This can translate into a security breach/issue for us.

There are several bugs that were fixed between 2.7.3 and 2.7.10 (current) that touch on `os.path.normpath` or `ntpath.normpath`.

Changes
=======

Since we are already using posix path separator internally for building paths I've replaced the calls to `os.path.normpath` with `posixpath.normpath`.

I've also extended the existing tests for getSegments to cover the issue.

reviewers: @adiroiban 

Please check that changes make sense.

### DIFF
--- a/chevah/compat/tests/normal/test_filesystem.py
+++ b/chevah/compat/tests/normal/test_filesystem.py
@@ -38,6 +38,65 @@ class FilesystemTestMixin(object):
             self.addCleanup(mk.fs.deleteFile, link_segments)
         return link_segments
 
+    def test_getSegments_upper_paths(self):
+        """
+        It will properly remove parent folder (..) and root folder (.) in
+        relative and absolute paths.
+        """
+        segments = self.filesystem.getSegments(u'../a/b')
+        self.assertEqual([u'a', u'b'], segments)
+
+        segments = self.filesystem.getSegments(u'/../a/b')
+        self.assertEqual([u'a', u'b'], segments)
+
+        segments = self.filesystem.getSegments(u'//../a/b')
+        self.assertEqual([u'a', u'b'], segments)
+
+        segments = self.filesystem.getSegments(u'./a/b')
+        self.assertEqual([u'a', u'b'], segments)
+
+        segments = self.filesystem.getSegments(u'/./a/b')
+        self.assertEqual([u'a', u'b'], segments)
+
+        segments = self.filesystem.getSegments(u'//./a/b')
+        self.assertEqual([u'a', u'b'], segments)
+
+    def test_getPath_empty(self):
+        """
+        It will return `/` when segments are empty.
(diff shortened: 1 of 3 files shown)


alibotean:
As a side-note: the failing test on server passes with the new changes in place.

alibotean:
Tests pass but I had a failure on `rhel6`: https://buildbot.chevah.com:10443/builders/compat-rhel-6/builds/76

On second run it worked.

alibotean:
needs-review

adiroiban:
The new tests look very similar. Maybe they can be shared.

These compat tests are a bad example of how a test should be written.
Please consider adding some comments to explain what is tested there.

Please also add tests to getPath so that we have 100% coverage for this patch.
Soon I plan to update PQM to not land if not all tests have pass that is buildbot + coverage.
 
You can install this extension and you should get a nice view of coverage.

I think that we also need to release a new compat version.

Please update the code so that all 3 tests pass and we have all green in GitHub

changes-approved

Thanks!

alibotean:
Thanks for the review I've added tests for `getPath` -- my fault for missing that change.

All is green but:

https://github.com/chevah/compat/pull/62/files#diff-3e24b47b27fa4074ab24117885532247R92

this looks very weird to me. I've adapted the test to fit the existing code though.

I will add a fixme if you confirm this is not intended behaviour.

alibotean:
The tests are reused as well; there was already a Mixin. Please note that I've added an alias `filesystem` for locked/unlocked to make my life easier.

needs-review

chevah/compat/tests/normal/test_filesystem.py
```diff
@@ -1381,37 +1441,22 @@ def test_getSegments(self):
...
+        segments = self.unlocked_filesystem.getSegments(u'/Aa/././bB')
+        self.assertEqual([u'Aa', u'bB'], segments)
+
         bubu_segments = home_segments[:]
         bubu_segments.append(u'Bubu')
         segments = self.unlocked_filesystem.getSegments(u'./Bubu')
         self.assertEqual(bubu_segments, segments)
 
-        # Going deep in the root will block at root folder.
+    def get_getSegments_deep_upper(self):
```
adiroiban: maybe `test_getSegments_deep_upper`

adiroiban:
Looks good. Please fix get_getSegments_deep_upper to have 100% patch coverage and merge.

Thanks!

changes-approved

PR #63, opened 2016-10-13
[#3646] Add API for writing with offset.

adiroiban:
Scope
=====

chevah.compat does not allow opening a file in write mode and writing at arbitrary locations.

This is useful for allowing arbitrary write access.

This is needed for implementing REST command for FTP

Changes
=======

Add an openFileForUpdating API

reviewers: @alibotean 

check that changes make sense

thanks!

### DIFF
--- a/chevah/compat/posix_filesystem.py
+++ b/chevah/compat/posix_filesystem.py
@@ -364,7 +364,8 @@ class PosixFilesystemBase(object):
                 fd = os.open(
                     path_encoded,
                     self.OPEN_READ_ONLY,
-                    DEFAULT_FILE_MODE)
+                    DEFAULT_FILE_MODE,
+                    )
                 return os.fdopen(fd, 'rb')
 
     def openFileForWriting(self, segments, utf8=False):
@@ -381,6 +382,32 @@ class PosixFilesystemBase(object):
                     path_encoded,
                     (self.OPEN_WRITE_ONLY | self.OPEN_CREATE |
                         self.OPEN_TRUNCATE),
+                    DEFAULT_FILE_MODE)
+                return os.fdopen(fd, 'wb')
+
+    def openFileForUpdating(self, segments, utf8=False):
+        """
+        See `ILocalFilesystem`.
+        """
+        path = self.getRealPathFromSegments(segments)
+        path_encoded = self.getEncodedPath(path)
+
+        self._requireFile(segments)
+        with self._IOToOSError(path), self._impersonateUser():
+            if utf8:
+                # This will also allow reading, but I hope we will
+                # remove this API soon, and we patch the high level API.
+                file = codecs.open(path_encoded, 'r+', 'utf-8')
+
+                def deny_read(size=None):
+                    raise IOError('File not open for reading')
+
+                file.read = deny_read
+                return file
+            else:
+                fd = os.open(
+                    path_encoded,
+                    self.OPEN_WRITE_ONLY,
                     DEFAULT_FILE_MODE)
                 return os.fdopen(fd, 'wb')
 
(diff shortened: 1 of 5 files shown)


adiroiban:
needs-review

adiroiban:
needs-review

chevah/compat/posix_filesystem.py
```diff
@@ -384,6 +385,32 @@ def openFileForWriting(self, segments, utf8=False):
...
+                    raise IOError('File not open for reading')
+
+                file.read = deny_read
+                return file
+            else:
+                fd = os.open(
+                    path_encoded,
+                    self.OPEN_WRITE_ONLY,
+                    DEFAULT_FILE_MODE)
+                return os.fdopen(fd, 'wb')
```
alibotean: I notice that we use the following modes: `r+` when supporting UTF8 and `wb` for the rest. The `os.fdopen` API doc sais that the modes are the same as for `open` which eventually `codecs.open` calls.

* ``r+''  Open for reading and writing.  The stream is positioned at the beginning of the file.
* ``w+''  Open for reading and writing.  The file is created if it does not exist, otherwise it is truncated.  The stream is positioned at  the beginning of the file.

I see that the position is set at the beginning on both cases but the second one mentions it will truncate the file. 

I will double check if we have a test for both variants (UTF8/other) and that we check for truncation.

chevah/compat/tests/normal/test_filesystem.py
```diff
@@ -1356,6 +1376,114 @@ def test_openFileForWriting_truncate(self):
         file_content = mk.fs.getFileContent(self.test_segments)
         self.assertEqual(new_content, file_content)
 
+    def test_openFileForUpdating_non_existing(self):
+        """
+        An error is raised when trying to open a file which does not
+        exists for updating..
```
alibotean: Double .

chevah/compat/tests/normal/test_filesystem.py
```diff
@@ -1356,6 +1376,114 @@ def test_openFileForWriting_truncate(self):
...
+        When a file is opened for updating the previous content is kept and
+        it will allow writing at arbitrary offsets.
+        """
+        content = b'some existing content'
+        new_content = b'more here'
+        # Create initial content.
+        self.test_segments = mk.fs.createFileInTemp(content=content)
+
+        # Write new content into file.
+        test_file = self.filesystem.openFileForUpdating(
```
alibotean: I see we have a test and it works as advertised on our part, now this is confusing. I will do some tests on my machine as well.

chevah/compat/posix_filesystem.py
alibotean: There must be a problem with the documentation I've found. As opening it for write only it will not truncate the file. Tested on Python 2.7.12 and Ubuntu.

adiroiban: Please note that for the os.fdopen(fd, 'wb') the FD is already created ... so the flags are just a helper for the Python File Object to know which operation to reject right away and to not even try to touch the low level OS api call.

alibotean: > Please note that for the os.fdopen(fd, 'wb') the FD is already created

I noticed that as well but the descriptor is obtained via `open(path, OPEN_WRITE_ONLY)` which according to the documentation found will truncate.

My opinion is that the documentation is not ok; it's from a post on Stack Exchange anyway. When testing locally I used `os.open` with `O_WRONLY`.

adiroiban: OPEN_WRITE_ONLY is our alias for `os.O_WRONLY | os.O_BINARY` 

As long as `os.O_TRUNC` is not used, the file should not be truncated

alibotean:
changes-approved

PR #64, opened 2016-12-02
[#2234] Check for SeImpersonatePrivilege when reporting Windows capabilities.

alibotean:
Scope
=====

When running SFTPPlus in debug without administrator privileges it's unable to authenticate OS accounts.

The reason for this is that without `SeImpersonatePrivilege` we cannot impersonate the user and access it's home folder.

For legacy reasons on Windows we were reporting this availability as always available.

Changes
=======

Fixed the code to check for the specified privilege when reporting local account impersonation capability.

reviewers: @adiroiban 

Please check that changes make sense. Once this is in I will fix the documentation on the server to describe this scenario and what should be done.

### DIFF
--- a/chevah/compat/tests/elevated/test_system_users.py
+++ b/chevah/compat/tests/elevated/test_system_users.py
@@ -423,9 +423,9 @@ class TestSystemUsers(SystemUsersTestCase):
             self.assertEqual(TEST_ACCOUNT_GROUP, impersonated_groupname)
             self.assertNotEqual(initial_uid, uid)
             self.assertNotEqual(initial_gid, gid)
-            # FIXME:3808:
-            # Investigate why this no longer works/passes on OSX.
             if self.os_name != 'osx':
+                # FIXME:3808:
+                # Investigate why this no longer works/passes on OSX.
                 self.assertNotEqual(initial_groups, impersonated_groups)
                 # On OSX newer than 10.5 get/set groups are useless.
                 self.assertEqual(
--- a/chevah/compat/tests/elevated/test_system_users.py
+++ b/chevah/compat/tests/elevated/test_system_users.py
@@ -423,8 +423,10 @@ class TestSystemUsers(SystemUsersTestCase):
             self.assertEqual(TEST_ACCOUNT_GROUP, impersonated_groupname)
             self.assertNotEqual(initial_uid, uid)
             self.assertNotEqual(initial_gid, gid)
-            self.assertNotEqual(initial_groups, impersonated_groups)
+            # FIXME:3808:
+            # Investigate why this no longer works/passes on OSX.
             if self.os_name != 'osx':
+                self.assertNotEqual(initial_groups, impersonated_groups)
                 # On OSX newer than 10.5 get/set groups are useless.
                 self.assertEqual(
                     sorted(self.getGroupsIDForTestAccount()),
--- a/chevah/compat/tests/normal/test_capabilities.py
+++ b/chevah/compat/tests/normal/test_capabilities.py
@@ -43,8 +43,17 @@ class TestProcessCapabilities(CompatTestCase):
         # 2003 and XP under normal mode.
         if 'win-2003' in self.hostname or 'win-xp' in self.hostname:
             return False
-        else:
-            return True
+
+        return True
+
+    def hasUAC(self):
+        """
+        Return True if Windows version has is UAC capable, False otherwise.
+        """
+        if 'win-2003' in self.hostname or 'win-xp' in self.hostname:
+            return False
+
+        return True
 
     def test_init(self):
         """
@@ -68,13 +77,16 @@ class TestProcessCapabilities(CompatTestCase):
         administration mode and disabled for the rest.
 
         See: runningAsAdministrator
+        See: hasUAC
         """
         result = self.capabilities.impersonate_local_account
 
         if self.runningAsAdministrator():
             self.assertTrue(result)
-        else:
+        elif self.hasUAC():
             self.assertFalse(result)
+        else:
+            self.assertTrue(result)
 
     def test_create_home_folder(self):
         """
--- a/chevah/compat/tests/normal/test_capabilities.py
+++ b/chevah/compat/tests/normal/test_capabilities.py
@@ -18,7 +18,11 @@ from zope.interface.verify import verifyObject
 from chevah.compat import process_capabilities
 from chevah.compat.exceptions import AdjustPrivilegeException
 from chevah.compat.interfaces import IProcessCapabilities
-from chevah.compat.testing import CompatTestCase, mk
+from chevah.compat.testing import (
+    CompatTestCase,
+    mk,
+    conditionals,
+    )
 
 
 class TestProcessCapabilities(CompatTestCase):
@@ -48,18 +52,29 @@ class TestProcessCapabilities(CompatTestCase):
         """
         verifyObject(IProcessCapabilities, self.capabilities)
 
-    def test_impersonate_local_account(self):
+    @conditionals.onOSFamily('posix')
+    def test_impersonate_local_account_posix(self):
         """
         When running under normal account, impersonation is always False
-        on Unix and always True on Windows.
+        on Unix.
         """
         result = self.capabilities.impersonate_local_account
-        if os.name == 'posix':
-            self.assertFalse(result)
-        elif os.name == 'nt':
+        self.assertFalse(result)
+
+    @conditionals.onOSFamily('nt')
+    def test_impersonate_local_account_windows(self):
+        """
+        Impersonation is enabled on machines running the tests in administration
+        mode and disabled for the rest.
+
+        See: runningAsAdministrator
+        """
+        result = self.capabilities.impersonate_local_account
+
+        if self.runningAsAdministrator():
             self.assertTrue(result)
         else:
-            raise AssertionError('Unsupported os.')
+            self.assertFalse(result)
 
     def test_create_home_folder(self):
         """
--- a/chevah/compat/tests/normal/test_capabilities.py
+++ b/chevah/compat/tests/normal/test_capabilities.py
@@ -72,8 +72,8 @@ class TestProcessCapabilities(CompatTestCase):
         # Windows 2003 BS is configured to execute with a non admin user.
         if 'win-2003' in self.hostname:
             self.assertFalse(result)
-
-        self.assertTrue(result)
+        else:
+            self.assertTrue(result)
 
     def test_create_home_folder(self):
         """
@@ -110,19 +110,20 @@ class TestProcessCapabilities(CompatTestCase):
         Check getCurrentPrivilegesDescription.
         """
         text = self.capabilities.getCurrentPrivilegesDescription()
+
         if self.os_family == 'posix':
             self.assertEqual(u'root capabilities disabled.', text)
         else:
-            # Normal Window account can impersonate.
             self.assertContains('SeChangeNotifyPrivilege:3', text)
-            self.assertContains('SeImpersonatePrivilege:3', text)
-            self.assertContains('SeCreateGlobalPrivilege:3', text)
 
             if self.runningAsAdministrator():
                 self.assertContains('SeCreateSymbolicLinkPrivilege:0', text)
+                self.assertContains('SeImpersonatePrivilege:3', text)
+                self.assertContains('SeCreateGlobalPrivilege:3', text)
             else:
-                self.assertNotContains(
-                    'SeCreateSymbolicLinkPrivilege', text)
+                self.assertNotContains('SeCreateSymbolicLinkPrivilege', text)
+                self.assertNotContains('SeImpersonatePrivilege', text)
+                self.assertNotContains('SeCreateGlobalPrivilege', text)
 
     def test_pam(self):
         """
@@ -163,15 +164,16 @@ class TestNTProcessCapabilities(TestProcessCapabilities):
         result = self.capabilities._getAvailablePrivileges()
 
         self.assertIsNotEmpty(result)
-        privilege = self.capabilities._getPrivilegeID(
-            win32security.SE_IMPERSONATE_NAME)
-        self.assertContains((privilege, 3), result)
-
-        privilege = self.capabilities._getPrivilegeID(
-            win32security.SE_SECURITY_NAME)
-        self.assertContains((privilege, 0), result)
 
         if self.runningAsAdministrator():
+            privilege = self.capabilities._getPrivilegeID(
+                win32security.SE_SECURITY_NAME)
+            self.assertContains((privilege, 0), result)
+
+            privilege = self.capabilities._getPrivilegeID(
+                win32security.SE_IMPERSONATE_NAME)
+            self.assertContains((privilege, 3), result)
+
             privilege = self.capabilities._getPrivilegeID(
                 win32security.SE_CREATE_SYMBOLIC_LINK_NAME)
             self.assertContains((privilege, 0), result)
@@ -204,7 +206,10 @@ class TestNTProcessCapabilities(TestProcessCapabilities):
         result = self.capabilities._getPrivilegeState(
             win32security.SE_SECURITY_NAME)
 
-        self.assertEqual(u'present', result)
+        if self.runningAsAdministrator():
+            self.assertEqual(u'present', result)
+        else:
+            self.assertEqual(u'absent', result)
 
     def test_getPrivilegeState_enabled_default(self):
         """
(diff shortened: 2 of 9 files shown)


alibotean:
needs-review

adiroiban:
looks good. we need to see what buildslave is missing or what test is missing to get these changes to 100%

alibotean:
> I don't see any new tests to support the new if branch

Yeap, there were and there are no tests for `impersonate_local_account`. Checked before sending this for review.

I will look into adding both available/not available functional tests.

alibotean:
> a non-mocked test would be best... maybe on win2003

I've added the test, so far we have:

* elevated test that is run on all BSes and checks for impersonation to be `True`
* normal test run on posix BSes that checks impersonation is `False`
* normal test run on Win 2003 / XP (don't we have it anymore) that checks impersonation is `False`
* normal test (same as above) run on all other Windows BSes that checks impersonation is `True`

I think we're covered.

adiroiban:
thanks. assuming that the tests will pass... changes-approved

chevah/compat/tests/normal/test_capabilities.py
```diff
@@ -43,8 +43,17 @@ def runningAsAdministrator(self):
         # 2003 and XP under normal mode.
         if 'win-2003' in self.hostname or 'win-xp' in self.hostname:
             return False
-        else:
-            return True
+
+        return True
```
adiroiban: this is fine, but in the future I think that we should avoid doing edit wars, in which the code is reformatted without changing the logic :)

chevah/compat/tests/normal/test_capabilities.py
```diff
@@ -43,8 +43,17 @@ def runningAsAdministrator(self):
         # 2003 and XP under normal mode.
         if 'win-2003' in self.hostname or 'win-xp' in self.hostname:
```
adiroiban: maybe we should try this

http://stackoverflow.com/a/1026626/539264

chevah/compat/tests/normal/test_capabilities.py
```diff
@@ -43,8 +43,17 @@ def runningAsAdministrator(self):
...
-        else:
-            return True
+
+        return True
+
+    def hasUAC(self):
+        """
+        Return True if Windows version has is UAC capable, False otherwise.
+        """
+        if 'win-2003' in self.hostname or 'win-xp' in self.hostname:
```
adiroiban: I think that for this we can do better.
We know that UAC is availalble from Vista and above.

maybe we can use something based on platform.win32_ver()[1]

I get:

* 2003 - > 5.7.3790
* 2008 -> 6.1.7000

so I think that we can use this value

chevah/compat/tests/normal/test_capabilities.py
```diff
@@ -68,13 +77,16 @@ def test_impersonate_local_account_windows(self):
         administration mode and disabled for the rest.
 
         See: runningAsAdministrator
+        See: hasUAC
```
adiroiban: not sure what to say... we are using this method only in a single test... maybe we should make it an embedded helper method to make it closer to the test code and so that we don't need this `See: hasUAC` docststring... this docstring feels strage

adiroiban:
thanks. Changes looks good, but I don't like the hasUAC method

please see the inline comments

needs-changes

alibotean:
I've tried the `ctypes` method but it does not work for us, a lot of tests are failing. My manual tests on Win 2003 looked promising.

I removed the `hasUAC` method along with my dependency on `runningAsAdministrator` as we do not yet have this combination and they were making the test hard to read/understand.

Apart from OSX, on which there is a deps issue. All the BSes are passing.

alibotean:
I've made the same changes to pavement, to install all dependencies in a single step but that is not the cause:

```
ImportError: No module named paver.tasks
```

chevah/compat/tests/normal/test_capabilities.py
```diff
@@ -39,21 +39,14 @@ def runningAsAdministrator(self):
...
         """
-        # Windows 2008 and DC client tests are done in administration mode,
-        # 2003 and XP under normal mode.
-        if 'win-2003' in self.hostname or 'win-xp' in self.hostname:
-            return False
+        admin = False
+        try:
+            admin = os.getuid() == 0
+        except AttributeError:
+            import ctypes
```
adiroiban: I think that this ctypes import can be top level, as all our python versions should have ctypes available

adiroiban:
on osx there is a separate error see the log
```
pip can't proceed with requirement 'chevah-brink==0.55.25' due to a pre-existing build directory.
 location: /private/var/folders/t4/31my8xv570x2jdt2nb6449v40000gp/T/pip_build_buildslave/chevah-brink
This is likely due to a previous installation that failed.
pip is being responsible and not assuming it can delete this.
Please delete it and try again.
```

you need to delete the file from the slave and try again... you should have ssh to that slave

alibotean:
I've added a fixme for checking what is going on/happened on OSX. Looking at the coverage report we do not have a machine where the tests do not have impersonation capabilities.

needs-review

adiroiban:
win-2003 is now failing as it now runs under a separate environment.

please take a look and hope we can get all the 3 checks green.

thanks

needs-changes

alibotean:
> please take a look and hope we can get all the 3 checks green.

Done. The tests fail for XP as `buildslave` is an administrator. Instead of making a mess of the tests I suggest we do the same for XP as well. With this change it will finally be as stated in the docstring.comments of `runningAsAdministrator`. I took the liberty of changing it to a regular/limited user myself -- only the BS process needs to be restarted if you agree.

Code coverage for the patch is 100% which should be comforting.

needs-review

adiroiban:
> With this change it will finally be as stated in the docstring/comments of runningAsAdministrator.

There  are 2 implemenation.

chevah.compat.testing.CompatTestCase.runningAsAdministrator 
chevah.compat.tests.normal.test_capabilities.TestProcessCapabilities.runningAsAdministrator

I think that we should remove the one from compat.testing.

-------

Please restart the XP slave

chevah/compat/tests/normal/test_capabilities.py
```diff
@@ -48,18 +52,28 @@ def test_init(self):
...
         """
         result = self.capabilities.impersonate_local_account
-        if os.name == 'posix':
+        self.assertFalse(result)
+
+    @conditionals.onOSFamily('nt')
+    def test_impersonate_local_account_windows(self):
+        """
+        Impersonation is available on all our Windows machines except Windows
+        2003.
```
adiroiban: not sure if this docstring is ok.

maybe ...

```
On Windows we have various slaves some running as administrators and they can impersonate.
```

adiroiban:
> With this change it will finally be as stated in the docstring/comments of runningAsAdministrator.

I also think that this should be renamed to runningAsWindowsAdministrator as from what I can see it is only used in the Windows context

chevah/compat/tests/elevated/test_system_users.py
```diff
@@ -423,8 +423,10 @@ def test_executeAsUser_Unix(self):
             self.assertEqual(TEST_ACCOUNT_GROUP, impersonated_groupname)
             self.assertNotEqual(initial_uid, uid)
             self.assertNotEqual(initial_gid, gid)
-            self.assertNotEqual(initial_groups, impersonated_groups)
+            # FIXME:3808:
+            # Investigate why this no longer works/passes on OSX.
```
adiroiban: not sure if this should be separate ticket... this is still related to set/get groups

chevah/compat/tests/normal/test_capabilities.py
```diff
@@ -149,16 +164,17 @@ def test_getAvailablePrivileges(self):
...
-            win32security.SE_IMPERSONATE_NAME)
-        self.assertContains((privilege, 3), result)
-
-        privilege = self.capabilities._getPrivilegeID(
-            win32security.SE_SECURITY_NAME)
-        self.assertContains((privilege, 0), result)
 
         if self.runningAsAdministrator():
             privilege = self.capabilities._getPrivilegeID(
+                win32security.SE_SECURITY_NAME)
```
adiroiban: I would say that this test should be split into test_getAvailablePrivileges_norma_user and test_getAvailablePrivileges_administrator

adiroiban:
To clean up the tests I think that we have

TestNTProcessCapabilitiesNormalUser and TestNTProcessCapabilitiesAdministrator

I feel that there are to many skips and conditinals in the current TestNTProcessCapabilities tests

Thanks!

needs-changes

alibotean:
> To clean up the tests I think that we have
> TestNTProcessCapabilitiesNormalUser and TestNTProcessCapabilitiesAdministrator

Ok

chevah/compat/tests/elevated/test_system_users.py
alibotean: I'm against fixing this here for the following reasons: I don't know what is causing this but master is also affected; I prefer to have a clean PR and not mix group related bugs with impersonation.

alibotean:
I've split all the tests into the following categories:
* TestProcessCapabilitiesPosix - will execute only on Linux/Unix based OSes
* TestNTProcessCapabilities - will execute on all Windows OSes
* TestNTProcessCapabilitiesNormalUser -- will execute only in Win 2003 and Win XP
* TestNTProcessCapabilitiesAdministrator - will execute on all other Windows machines

We do not have 100% coverage because there is no HP-UX buildslave at the moment: https://codecov.io/gh/chevah/compat/compare/5bb0cd4413b4f627d14ce44f87942af49c70669b...e0edb7c99b89d9f0a753c53d57cc286cdcc4b2cb#6368657661682F636F6D7061742F74657374732F6E6F726D616C2F746573745F6361706162696C69746965732E7079-84

alibotean:
needs-review

alibotean:
I don't know how to fix the coverage for this:

https://codecov.io/gh/chevah/compat/compare/5bb0cd4413b4f627d14ce44f87942af49c70669b...ed62651f4fd61fa290a6fa04a76d2cd10fca2262#6368657661682F636F6D7061742F74657374732F6E6F726D616C2F746573745F6361706162696C69746965732E7079-148

chevah/compat/tests/normal/test_capabilities.py
```diff
@@ -51,80 +45,60 @@ def test_init(self):
...
+        self.assertEqual(u'root capabilities disabled.', text)
 
     def test_pam(self):
         """
         PAM is supported on Linux/Unix with the exception of HPUX.
+        """
+        # FIXME:3813:
+        # Re-enable once we have a HP-UX buildslave.
+        # if self.os_name in ['hpux']:
+        #    self.assertFalse(self.capabilities.pam)
```
adiroiban: just remove this code and provide the required details in the 3814 ticket.

there is no reason to keep commented code in the repo

please note that 3813 is about setting up the hp-ux buildslave ... but we need separate tickets to fix hp-ux support on compat / server ... etc

there is also 2703 which talks about creating a python-package for HP-UX https://github.com/chevah/python-package/pull/25

chevah/compat/tests/normal/test_capabilities.py
```diff
@@ -142,25 +124,75 @@ def test_openProcess_success(self):
...
+                    )):
+                # pragma: no cover
+                pass
+
+    def test_get_home_folder(self):
+        """
+        On Windows, only Windows 2008 and Windows 7 can get home folder path.
+        """
+        # The Windows test is handled in elevated module.
+        pass
```
adiroiban: I don't understand this. Why not write some testing code here?

chevah/compat/tests/normal/test_capabilities.py
```diff
@@ -218,17 +251,121 @@ def test_isPrivilegeEnabled_disabled(self):
...
+        self.assertFalse(result)
+
+    def test_create_home_folder(self):
+        """
+        On Windows home folders can be created if required privileges
+        are configured for the process.
+        """
+        result = self.capabilities.create_home_folder
+
+        if 'win-xp' in self.hostname:
```
adiroiban: I think that this need a comment to explain why XP is different

chevah/compat/tests/normal/test_capabilities.py
```diff
@@ -218,17 +251,121 @@ def test_isPrivilegeEnabled_disabled(self):
...
+    def test_getCurrentPrivilegesDescription(self):
+        """
+        Check getCurrentPrivilegesDescription.
+        """
+        text = self.capabilities.getCurrentPrivilegesDescription()
+
+        self.assertContains('SeChangeNotifyPrivilege:3', text)
+
+        if 'win-xp' in self.hostname:
+            self.assertNotContains('SeCreateSymbolicLinkPrivilege', text)
```
adiroiban: we need a comment to explain why we have this branch.

doing assertions inside an if branch is a bad thing to do in testing... usually a smell that the test should be split... but if we don't split it, then we need a comment as a placeholder for the docstring which would have been associated with the splitted test.

chevah/compat/tests/normal/test_capabilities.py
```diff
@@ -218,17 +251,121 @@ def test_isPrivilegeEnabled_disabled(self):
...
+        super(TestNTProcessCapabilitiesAdministrator, self).setUp()
+
+        if self.os_family != 'nt':
+            raise self.skipTest("Only Windows platforms supported.")
+
+        # Windows 2008 and DC client tests are done in administration mode,
+        # 2003 and XP under normal mode.
+        if 'win-2003' in self.hostname or 'win-xp' in self.hostname:
+            message = "Only Windows with administrator rights supported"
+            raise self.skipTest(message)
```
adiroiban: this whole dance is ok, but it would be nice to have it as an skipOnCondition decorator.

the main reason as that with this code we will still execute the setUp code for each test, which if we have a skipOnCondition for the whole testcase the setUp code will never be executed

chevah/compat/tests/normal/test_capabilities.py
```diff
@@ -328,17 +454,29 @@ def test_elevatePrivilege_multiple_privileges_success(self):
...
+        On Windows home folders can be created if required privileges
+        are configured for the process.
+        """
+        result = self.capabilities.create_home_folder
+
+        self.assertTrue(result)
+
+    def test_getCurrentPrivilegesDescription(self):
+        """
+        Check getCurrentPrivilegesDescription.
```
adiroiban: can we have a better docstring here :)

chevah/compat/tests/normal/test_capabilities.py
```diff
@@ -328,17 +454,29 @@ def test_elevatePrivilege_multiple_privileges_success(self):
...
-            return
+        self.assertTrue(symbolic_link)
 
-        if self.runningAsAdministrator():
-            self.assertTrue(symbolic_link)
-        else:
-            self.assertFalse(symbolic_link)
+    def test_create_home_folder(self):
+        """
+        On Windows home folders can be created if required privileges
```
adiroiban: instead of "required privileges" can we be more explicit and add details about which exact privileges are requried... like Run As Admin, SE_BACKUP...etc

adiroiban:
Changes looks good, but if possible instead of the setUp skip I would prefer to either have a setUpClass skip or best a `@conditionals.skipOnCondition` decorator to skip the whole testcase and to not execute setUpCllass or part of setUp

I left other inline comments.

thanks!

needs-changes

adiroiban:
please update with master

chevah/compat/tests/normal/test_capabilities.py
alibotean: I don't know, that's what the test said originally. I will look into what is going on.

chevah/compat/tests/normal/test_capabilities.py
alibotean: We have it now; I will replace all these.

alibotean:
I think I've ticked all the boxes. Tests and coverage looking good.

needs-review

chevah/compat/tests/elevated/test_system_users.py
```diff
@@ -423,10 +423,11 @@ def test_executeAsUser_Unix(self):
             self.assertEqual(TEST_ACCOUNT_GROUP, impersonated_groupname)
             self.assertNotEqual(initial_uid, uid)
             self.assertNotEqual(initial_gid, gid)
-
+            # FIXME:3808:
+            # Investigate why this no longer works/passes on OSX.
```
adiroiban: I think that this comment should be moved inside the block

pavement.py
```diff
@@ -152,37 +152,15 @@ def update_setup():
...
-    Install dependencies for build environment.
-    """
-    print('Installing build dependencies to %s...' % (pave.path.build))
+    print('Installing dependencies to %s...' % (pave.path.build,))
     pave.pip(
         command='install',
-        arguments=BUILD_PACKAGES,
-        silent=True,
+        arguments=RUN_PACKAGES + TEST_PACKAGES + BUILD_PACKAGES,
+        silent=False,
```
adiroiban: I think that we can just remote the silent argument and go with the default

adiroiban:
changes look good.

changes-approved

PR #65, opened 2017-01-23
[#3817] Add empirical to compat.

adiroiban:
Scope
=====

This moves the empirical code into compat

empirical depends on compat, and compat depends of empirical so they should be a single package.

Changes
=======

For now I have just moved the code... and chevah.empirical is in its own namespace

The empirical tests were moved to chevah.compat so that they were executed as our build system don't support multi-namespace repos.

The plan is to move chevah.empirical into chevah.compat.testing ... but this in a separate PR

reviewers: @alibotean 

check that changes make sense


alibotean:
Let's do this 👍 

changes-approved

PR #67, opened 2017-01-26
[#3830] Move chevah.emprical to chevah.compat.testing

adiroiban:
Scope
=====

The fact that in previous version chevah-compat package was installing both chevah.empirical and chevah.compat namespaces was creating a conflict with the existing chevah.empirical package.

Changes
=======

Moved all empirical to chevah.compat.testing and reorganized the chevah/compat/testing.py 

reviewers: @alibotean 

This will need to be tested as part of the work for 3828


adiroiban:
I think that the branch looks good.

I have just moved the code around to avoid circular references.

Since the code was moved it is now part of the coverage report... and with this I found some code which was not used

needs-review

alibotean:
> the patch might not be 100% cover...but I will try to improve the coverage as much as possible

Oki doh.

changes-approved

PR #68, opened 2017-01-27
[#3831] Make mockup compatible with py2.

adiroiban:
Scope
=====

The initial porting to py3 resulted in `mk` creating types from the future lib which broke our tests.

the new types from future are much better as they provide a strict interpretation of str vs bytes... but they break the chevah/server tests... so for now we make them backward compatible and after that we will update the chevah/server tests to work with the new types

Changes
=======

Update the code to use the `native` types.

I have removed the command test case as it was only used by the server code

reviewers: @alibotean 

check that changes make sense

### DIFF
--- a/chevah/compat/testing/mockup.py
+++ b/chevah/compat/testing/mockup.py
@@ -434,8 +434,11 @@ class ResponseDefinition(object):
         """
         Will update the content returned to the server.
         """
+        if not isinstance(content, bytes):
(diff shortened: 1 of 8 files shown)


adiroiban:
test pass on my local branch of chevah/server ... so I will push this for an initial release

needs-review

alibotean:
changes-approved

PR #69, opened 2017-02-09
[#3860] Use default timeout from test case

adiroiban:
Scope
=====

This defines the default timeout for a deferred in the test case.

In this way you can increase the timeout for all tests inside a slow testcase... and we don't have to pass it to all call.

Changes
=======

Define and use deferred timeout as instance member.

reviewers: @alibotean 

check that changes make sense

with this, we can update ServerTestCase with a default of 1.5 ... to see if we still have many tests fainling at random time due to slow deferred

### DIFF
--- a/chevah/compat/tests/normal/testing/test_testcase.py
+++ b/chevah/compat/tests/normal/testing/test_testcase.py
@@ -66,9 +66,9 @@ class TestTwistedTestCase(ChevahTestCase):
         self.assertEqual(
             'This is not a deferred.', context.exception.args[0])
 
-    def test_runDeferred_timeout(self):
+    def test_runDeferred_timeout_custom(self):
         """
-        runDeferred will execute the reactor and raise a timeout
+        runDeferred will execute the reactor and raise an exception
         if deferred got no result after the timeout.
         """
         deferred = defer.Deferred()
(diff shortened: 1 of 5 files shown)


adiroiban:
needs-review

alibotean:
Took the liberty to change the code a bit and add a couple of tests.

I'm for renaming the `DEFAULT_TIMEOUT` in all the server tests to more accurately reflect the timeout in question:

* `DB_TIMEOUT`
* `EMAIL_TIMEOUT`
* etc.

once we use this new compat version as we will have `DEFAULT_TIMEOUT` and `DEFERRED_TIMEOUT` but we use the first for deferreds.

changes-approved

chevah/compat/tests/normal/testing/test_testcase.py
```diff
@@ -81,7 +81,27 @@ def test_runDeferred_timeout(self):
...
+        # Restore order messing with internal timeout state in
+        # previous state.
+        self._reactor_timeout_failure = None
+
+    def test_runDeferred_timeout_default(self):
+        """
+        It will execute the reactor and raise an exception if the
+        default timeout passes and the deferred is not completed.
+        """
+        self.DEFERRED_TIMEOUT = 0
```
adiroiban: I think that we can go for 0.01 ... compat tests are fast

chevah/compat/tests/normal/testing/test_testcase.py
```diff
@@ -347,6 +367,43 @@ def much_later():  # pragma: no cover
...
+    def test_executeReactor_timeout_custom(self):
+        """
+        It will use the custom timeout value instead of the default,
+        test case value.
+        """
+        self.called = False
+
+        def last_call():
+            self.called = True
+        reactor.callLater(2, last_call)
```
adiroiban: maybe go for 2.5 :)

adiroiban:
> I'm for renaming the DEFAULT_TIMEOUT in all the server tests

yes. once this is merged I will create a separate cleanup ticket. 

thanks for the updated tests

PR #70, opened 2017-02-21
[#3880] Don't fail for already done delayed calls.

adiroiban:
Scope
=====

This fix the so that we don't see false errors in tearDown

Changes
=======

```
======================================================================
ERROR: test_get_non_empty_ascii_file - functional.test_ftp_transfer:TestDataChannelFTPSEActive.test_get_non_empty_ascii_file
----------------------------------------------------------------------
Traceback (most recent call last):
  File "c:/Users/domainbuildslave/chevah/bs/server-win-dc-client/build/chevah\server\testing\functional.py", line 163, in tearDown
    self.stopService()
  File "c:/Users/domainbuildslave/chevah/bs/server-win-dc-client/build/chevah\server\testing\functional.py", line 269, in stopService
    self.cleanReactor()
  File "c:\Users\domainbuildslave\chevah\bs\server-win-dc-client\build\build-windows-x86\lib\lib\site-packages\chevah\compat\testing\testcase.py", line 201, in cleanReactor
    delayed_call.cancel()
  File "c:\Users\domainbuildslave\chevah\bs\server-win-dc-client\build\build-windows-x86\lib\lib\site-packages\twisted\internet\base.py", line 89, in cancel
    raise error.AlreadyCalled
AlreadyCalled: Tried to cancel an already-called event.

```

I was not able reproduce this in a test...since this is a race condition due to threads but I wrote some kind of test to cover it.

reviewers: @brunogola @alibotean 

this is an example why `if self.canDoX(): self.doX()` is not working with threads ... and we should just do `try: self.doX() except`

### DIFF
--- a/chevah/compat/testing/testcase.py
+++ b/chevah/compat/testing/testcase.py
@@ -137,7 +137,7 @@ class TwistedTestCase(TestCase):
                 self.assertIsNone(self._reactor_timeout_failure)
                 self.assertReactorIsClean()
         finally:
-            self.cleanReactor()
+            self._cleanReactor()
         super(TwistedTestCase, self).tearDown()
 
     def _reactorQueueToString(self):
@@ -181,9 +181,12 @@ class TwistedTestCase(TestCase):
             return reactor.threadpool.working
 
     @classmethod
-    def cleanReactor(cls):
+    def _cleanReactor(cls):
         """
         Remove all delayed calls, readers and writers from the reactor.
+
+        This is only for cleanup purpose and should not be used by normal
+        tests.
         """
         if not reactor:
             return
--- a/chevah/compat/testing/testcase.py
+++ b/chevah/compat/testing/testcase.py
@@ -24,7 +24,6 @@ from bunch import Bunch
 from mock import patch, Mock
 from nose import SkipTest
 from twisted.internet.defer import Deferred
-from twisted.internet.error import AlreadyCalled, AlreadyCancelled
 from twisted.python.failure import Failure
 try:
     from twisted.internet.posixbase import (
--- a/chevah/compat/testing/testcase.py
+++ b/chevah/compat/testing/testcase.py
@@ -191,14 +191,21 @@ class TwistedTestCase(TestCase):
             reactor.removeAll()
         except (RuntimeError, KeyError):
             # FIXME:863:
-            # When running threads the reactor is cleaned from multiple places
-            # and removeAll will fail since it detects that internal state
+            # When running threads tests the reactor touched from the test
+            # case itself which run in one tread and from the fixtures/cleanup
+            # code which is executed from another thread.
+            # removeAll might fail since it detects that internal state
             # is changed from other source.
             pass
+
         reactor.threadCallQueue = []
         for delayed_call in reactor.getDelayedCalls():
-            if delayed_call.active():
+            try:
                 delayed_call.cancel()
+            except ValueError:
+                # AlreadyCancelled and AlreadyCalled are ValueError.
+                # Might be canceled from the separate thread.
+                pass
 
     def _raiseReactorTimeoutError(self, timeout):
         """
--- a/chevah/compat/testing/testcase.py
+++ b/chevah/compat/testing/testcase.py
@@ -23,12 +23,13 @@ import time
 from bunch import Bunch
 from mock import patch, Mock
 from nose import SkipTest
+from twisted.internet.defer import Deferred
+from twisted.internet.error import AlreadyCalled, AlreadyCancelled
+from twisted.python.failure import Failure
 try:
-    from twisted.internet.defer import Deferred
     from twisted.internet.posixbase import (
         _SocketWaker, _UnixWaker, _SIGCHLDWaker
         )
-    from twisted.python.failure import Failure
 except ImportError:
     # Twisted support is optional.
     _SocketWaker = None
@@ -152,7 +153,7 @@ class TwistedTestCase(TestCase):
 
     def _threadPoolQueueSize(self):
         """
-        Return current size of thread Pool, or None when treadpool does not
+        Return current size of thread Pool, or None when threadpool does not
         exists.
         """
         if not reactor.threadpool:
@@ -162,7 +163,7 @@ class TwistedTestCase(TestCase):
 
     def _threadPoolThreads(self):
         """
-        Return current threads from pool, or None when treadpool does not
+        Return current threads from pool, or None when threadpool does not
         exists.
         """
         if not reactor.threadpool:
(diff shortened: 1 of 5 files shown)


adiroiban:
should be good. needs-review

adiroiban:
in previous commits you can see that since there was a `lambda: None` function, the coverate was not 100% since that lamda function was never called.

chevah/compat/testing/testcase.py
```diff
@@ -181,24 +181,34 @@ def _threadPoolWorking(self):
...
             # is changed from other source.
             pass
+
         reactor.threadCallQueue = []
         for delayed_call in reactor.getDelayedCalls():
-            if delayed_call.active():
+            try:
                 delayed_call.cancel()
+            except ValueError:
+                # AlreadyCancelled and AlreadyCalled are ValueError.
```
alibotean: I'm in favor of catching `AlreadyCanceled` and `AlreadyCalled` explicitly.

chevah/compat/tests/normal/testing/test_testcase.py
```diff
@@ -380,6 +380,42 @@ def much_later():  # pragma: no cover
...
+        self.assertIsNotEmpty(reactor.getDelayedCalls())
+
+        self._cleanReactor()
+
+        self.assertIsEmpty(reactor.getDelayedCalls())
+
+    def test_cleanReactor_delayed_calls_some_called(self):
+        """
+        It  will not break if a call is already called and will continue
+        canceling the .
```
alibotean: unfinished docstring

chevah/compat/tests/normal/testing/test_testcase.py
```diff
@@ -380,6 +380,42 @@ def much_later():  # pragma: no cover
...
+        """
+        delayed_call_1 = reactor.callLater(1, self.ignoreFailure)
+        delayed_call_2 = reactor.callLater(2, self.ignoreFailure)
+        # Fake that deferred was called and make sure it is first in the list
+        # so that we can can check that the operation will continue.
+        delayed_call_1.called = True
```
alibotean: Just askin': why not use the `clock` and simply advance with 1 second so that it's called for real, then double check was called with an assert?

chevah/compat/tests/normal/testing/test_testcase.py
```diff
@@ -380,6 +380,42 @@ def much_later():  # pragma: no cover
...
+
+        self._cleanReactor()
+
+        self.assertTrue(delayed_call_2.cancelled)
+        # Since we are messing with the reactor, we are leaving it in an
+        # inconsistent state as no called delayed call should be part of the
+        # list... since when called, the delayed called is removed right
+        # away, yet we are not removing it but only faking its call.
+        delayed_call_1.called = False
+        self._cleanReactor()
```
alibotean: Will get rid of this ugly stuff as well.

alibotean:
We're sitting on a time bomb :smile:

```
TypeError: ignoreFailure() takes exactly 2 arguments (1 given)
```

alibotean:
Figured it our; we're using `Clock` which implements only `IReactorTime` and does not provide all the API used by `_cleanReactor`. 

It's doable if we go for `MemoryReactor` or `MemoryReactorClock`.

alibotean:
Please consider my comments.

changes-approved

alibotean:
I've fixed some small typos and changed the imports for Twisted -- please double check this -- it looked strange to consider Deferred optional.

chevah/compat/tests/normal/testing/test_testcase.py
adiroiban: this code is for the global reactor ... not for our mock clock

adiroiban:
@alibotean your changes are not ok... the idea is that Twisted support is optional and we should be able to use compat even with projects which are not using Twisted

-----

> We're sitting on a time bomb

Can you please add the details to reproduce this error? Thanks!

chevah/compat/tests/normal/testing/test_testcase.py
adiroiban: so when we use clock, there is no need to clean it... just check that the clock is clean and then just dispose here.... but the code here is for functional tests where we are using the real reactor, and in which we are re-using the same reactor ... and accessing the same reactor from a thread so that we can have the service in one thread and the testing bot in another thread

alibotean:
> the idea is that Twisted support is optional and we should be able to use compat even with projects which are not using Twisted

I see. My changes were in `testcase.py` only and by looking at the code all classes there use Twisted hence my assumption that even if Twisted is optional that is not the case for that module. There is no point to import the module without Twisted if all the code inside needs it anyway.

chevah/compat/tests/normal/testing/test_testcase.py
alibotean: > o when we use clock, there is no need to clean it...

Agreed, I've played with the test and the Clock and that was my design as well.

> but the code here is for functional tests where we are using the real reactor, and in which we are re-using the same reactor ... and accessing the same reactor from a thread so that we can have the service in one thread and the testing bot in another thread

I know, just that functional and mocking/hacking don't go well together and I was looking for a better way.

alibotean:
> Can you please add the details to reproduce this error? Thanks!

You need to simply let the delayed call be triggered and you will see the error -- the signature of `ignoreFailure` needs another parameter.

 But it depends on how you look at it -- since we actually don't want to execute it; it might better this way.

brunogola:
changes-approved

adiroiban:
> You need to simply let the delayed call be triggered and you will see the error -- the signature of ignoreFailure needs another parameter.

I am confused. Why would I need to let it trigger?

Can you please describe the docstring of such a test?

Thanks!

alibotean:
> Why would I need to let it trigger?

You don't need to let it trigger, here's the code from my experiment... you will understand:

```
    def test_cleanReactor_delayed_calls_some_called(self):
        """
        It will not break if a call is already called and will continue
        canceling the remaining calls.
        """
        delayed_call_1 = reactor.callLater(1, self.ignoreFailure)
        delayed_call_2 = reactor.callLater(2, self.ignoreFailure)
        self.assertEqual(
            [delayed_call_1, delayed_call_2], reactor.getDelayedCalls())
        reactor.advance(1.2)
        self.assertEqual(
            [delayed_call_2], reactor.getDelayedCalls())
        self.assertTrue(delayed_call1.called)

        self._cleanReactor()

        self.assertTrue(delayed_call_2.cancelled)
        self.assertFalse(delayed_call_2.called)
        self.assertIsEmpty(reactor.getDelayedCalls())
```

PR #71, opened 2017-03-08
[#1885] Update tests for HPUX.

adiroiban:
Scope
=====

We had a big gap of almost one year in which we did not had a hpux slave ... so now we have a new slave and we want to have a green build.

Changes
=======

I have just skipped the PAM tests as we have a separate ticket for that.

As a drive-by change I have removed the http context test code as it will be moved in server.

No release yet as there is not much API... but I have updated the version.
Once we will made another change we will release it together with the changes for testing code.

reviewers: @alibotean 

check that changes make sense

thanks!


adiroiban:
needs-review

adiroiban:
this is just skipping failing tests and removing a lot of code... I don't understand why there is a decrease in coverage

alibotean:
changes-approved

PR #72, opened 2017-04-24
[#3983] Add support for folder iterator API.

adiroiban:
Scope
=====

This is built on top of scandir module to allow a non-blocking / iterative way of listing a directory.

Changes
=======

the paver.sh and the whole build system was udpated to use the python binary archives based on versions.

Wrapped the scandir function to a compat like method.
It convers some Windows error to Unix one for simple handling.

As a drive by I have moved all the assertions into a separate code.

As a drive by the mk.bytes helper was updated to return a native bytes() and not the specific bytesarray object.

As a drive by, I have updated `paver deps` to not install the whole build dependecies on buildslaves, as this is not needed.

BUILD_PACKAGES are needed for PQM, test_review... etc, but the buildslaves are not doing this.... so there is no need to pull a lot of dependencies..
We can break this further and only have pylint/pyflakses dependencies installed on os-independent

I have created a few test  filesystem helpers which are creating folders with auto-cleanup.

reviewers: @alibotean @brunogola 

sorry for the long diff, but otherwise I would never done all this cleanup

thanks!


adiroiban:
hp-ux is still failing due to a bad python version, but I hope that @dumol will fix it soon.

for now, this is using the python-binary from testing... as I don't want to push the new python version in production until all is not ok... so not created issues with the server tests.

needs-review

adiroiban:
the coverage report is broken as for some reason we don't get the coverage from Windows. I am investigating this.

chevah/compat/posix_filesystem.py
```diff
@@ -36,12 +43,13 @@
 
 
 class PosixFilesystemBase(object):
-    '''Base implementation if ILocalFilesystem for
+    """
+    Base implementation if ILocalFilesystem for
```
alibotean: if -> of

chevah/compat/testing/assertion.py
```diff
@@ -0,0 +1,366 @@
+# -*- coding: utf-8 -*-
```
alibotean: Sice we are creating this anew (copy I know) it's a good time to place the `unicode_literals` import and revise the strings.

chevah/compat/testing/filesystem.py
```diff
@@ -160,6 +160,23 @@ def makePathInTemp(self):
         path = os.path.join(self.temp_path, name)
         return (path, segments)
 
+    def folder(self, segments, cleanup):
+        """
+        Create a folder and remove it a cleanup.
+        """
+        cleanup(self.deleteFolder, segments, recursive=True)
+        self.createFolder(segments)
```
alibotean: I prefer the below approach to add cleanup only on successful create.

chevah/compat/testing/mockup.py
```diff
@@ -358,6 +358,22 @@ def makeIPv4Address(self, host='localhost', port=None, protocol='TCP'):
         ipv4 = address.IPv4Address(protocol, host, port)
         return ipv4
 
+    def FilesystemOsAvatar(self, user, home_folder_path=None):
+        """
+        Create an avatar to by used with the test filesystem.
```
alibotean: by -> be

chevah/compat/testing/mockup.py
```diff
@@ -358,6 +358,22 @@ def makeIPv4Address(self, host='localhost', port=None, protocol='TCP'):
...
+
+        `user` is passed as a TestUser and all other
```
alibotean: Unfinished docstring

chevah/compat/tests/normal/testing/test_assertion.py
```diff
@@ -0,0 +1,188 @@
+# Copyright (c) 2015 Adi Roiban.
+# See LICENSE for details.
+"""
+Tests for the assertion helpers.
+"""
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
```
alibotean: Missing `unicode_literals` import

adiroiban: well... this is the code from imported from empirical, where instead of unicode_literals we went for `builtin str`

alibotean:
> This is built on top of scandir module to allow a non-blocking / iterative way of listing a directory.

I can't pinpoint a test or a couple of tests that "prove" the new approach is non blocking. I'm not saying it is not just that by reading the test I'm not convinced.

This would be the single one mentioning this:

https://github.com/chevah/compat/pull/72/files#diff-b756a25eaebd09f2a2ed5391b4697803R266

but the assert section is not different from others that do not claim this.

alibotean:
I think we can write a test to "prove" that the iteration returns immediately:



```
./touch aa
$ touch ab
$ touch ac
$ python
Python 2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:42:59) [MSC v.150
Type "help", "copyright", "credits" or "license" for more information.
>>> from _scandir import scandir
>>> from scandir import scandir_python as scandir
>>> k=scandir('.')
>>> print next(k)
<Win32DirEntryPython: 'aa'>
>>> import os
>>> os.remove('ab')
>>> os.remove('ab')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
WindowsError: [Error 2] The system cannot find the file specified: 'ab'
>>> print next(k)
<Win32DirEntryPython: 'ab'>
>>> print next(k)
<Win32DirEntryPython: 'ac'>
>>>                                                                   
```

alibotean:
needs-changes

chevah/compat/testing/assertion.py
adiroiban: even if we have unicode literals, we still need `# -*- coding: utf-8 -*-`

This diff is already big, and it was a bad idea to do these changes here... so I don't want to do more changes....

compat is already ported to py3 syntax

for compat we are not using unicode_literals, but the `future str`

adiroiban: still... assertions doesn't have any unicode text...so this can be removed

chevah/compat/testing/filesystem.py
adiroiban: true

adiroiban:
I wrote a test for checking that operation is not blocking.

the coverage is bad, but is due to the moved code.
I have check the coverage for the filesystem code and is 100%

for now aix is slow, so I made an exception for it

thanks!

needs-review

alibotean:
Now that's a heavy duty functional test :smile:

changes-approved

pavement.py
```diff
@@ -210,12 +231,27 @@ def test_ci(args):
...
+    default_args = [
+        '--with-run-reporter',
+        '--with-timer',
+        ]
+
+    # Show some info about the current environment.
+    from OpenSSL import SSL, __version__ as pyopenssl_version
+    from coverage.cmdline import main as coverage_main
+
+    print 'PYTHON %s' % (sys.version,)
```
brunogola: print syntax

adiroiban: yes... we need to update pocket-lint to work with pylint so that we include all the files in py3 tests.

but this is for another ticket

brunogola:
changes-approved

adiroiban:
test_iterateFolderContent_big is a PITA ... and we might need to tune it further or write it in another way to make it 100% reliable

thanks for the review

PR #73, opened 2017-05-05
[#4048] Revert spwd code. Fix build system

adiroiban:
Scope
=====

This fix a bad change in previous PR.

The coverage was reporting a bug missing chunk... but  I have just ignored it thinking that is old code... well, it was some old code, but it was also some unintentional code which was removed... and a lot of tests which were not executed.

We really need to add a hard check and make sure we have 100% coverage for the tests

Changes
=======

Add back the code which was accidentally removed.

Update the build system to execute the elevated test.

This add proper password auth support for freebsd and openbsd... with support for create users and groupd

The master branch was updated in GitHub to be protected from the default tests and the coverage/project/tests... if they are not green , you can't merge it

I made a few changes to the coverage reporting

reviewers: @brunogola 

check that changes make sense

I will try to update codecov to have a separate report for the coverage or the tests code itself

### DIFF
--- a/chevah/compat/administration.py
+++ b/chevah/compat/administration.py
@@ -46,7 +46,7 @@ def execute(command, input_text=None, output=None, ignore_errors=True):
     Execute a command having stdout redirected and using 'input_text' as
     input.
     """
-    verbose = False
+    verbose = True
 
     if verbose:
         print('Calling: %s' % command)
--- a/chevah/compat/administration.py
+++ b/chevah/compat/administration.py
@@ -41,13 +41,13 @@ from chevah.compat import (
 from chevah.compat.winerrors import ERROR_NONE_MAPPED
 
 
-def execute(
-    command, input_text=None, output=None, ignore_errors=True, verbose=False,
-        ):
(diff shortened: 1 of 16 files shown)


adiroiban:
much better now.. see the patch diff https://codecov.io/gh/chevah/compat/compare/edf45f5e4367948fd1eeca37955abb1b00a5941d...26586bb7230e1c9fddb2b8cc53358982098ba264/diff

also, we now have codecov/projects/tests which should trigger a big red merge button if is not 100%

needs-review

brunogola:
changes seem good.

changes-approved

chevah/compat/administration.py
```diff
@@ -41,19 +41,21 @@
...
 
-def execute(
-    command, input_text=None, output=None, ignore_errors=True, verbose=False,
-        ):
+def execute(command, input_text=None, output=None, ignore_errors=True):
     """
     Execute a command having stdout redirected and using 'input_text' as
     input.
     """
+    verbose = False
```
adiroiban: the verbose argument from execute was never used... this is more like an internal "debug" flag... which you enable it while investigating an issue

PR #74, opened 2017-06-23
[#4183] Update build system.

adiroiban:
Scope
=====

This updates the code to pass on all the buildslaves.

Changes
=======

I have fixed the py3 tests. I had to skip the scandir tests.

I have updated the linter, so I have updated the code to pass the static checks.

reviewers: @brunogola 

### DIFF
--- a/chevah/compat/tests/normal/testing/test_mockup.py
+++ b/chevah/compat/tests/normal/testing/test_mockup.py
@@ -60,6 +60,7 @@ class TestFactory(ChevahTestCase):
 
     def assertUnicodeDecodeError(self, exception):
         """
+        Check the error message on py2 and py3.
         """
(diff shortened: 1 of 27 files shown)


adiroiban:
I need to fix the raspbian-8 strange error, but should be ready for review

needs-review

chevah/compat/tests/normal/testing/test_mockup.py
```diff
@@ -57,6 +58,16 @@ def test_bytes(self):
         self.assertNotEqual(mk.bytes(), mk.bytes())
         self.assertIsInstance(bytes, mk.bytes())
 
+    def assertUnicodeDecodeError(self, exception):
+        """
```
brunogola: empty docstring

PR #76, opened 2017-09-05
[#4330] Add support for Alpine Linux.

adiroiban:
Scope
=====

This update the code to run on Alpine slave.

Changes
=======

Update paver.sh to recognize alpine.

For now PAM test is skipped and for osgroups we ignore duplicate groups.

As a drive-by change I have updated os_version detection in the test case to get the AIX version.

reviewers: @dumol 

Run running test on python-package you might want to use this branch instead of master.

### DIFF
--- a/chevah/compat/testing/filesystem.py
+++ b/chevah/compat/testing/filesystem.py
@@ -322,7 +322,7 @@ class LocalTestFilesystem(LocalFilesystem):
                 md5_sum.update(chunk)
         finally:
             input_file.close()
-        return md5_sum.digest()
+        return md5_sum.hexdigest()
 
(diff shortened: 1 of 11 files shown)


adiroiban:
needs-review

dumol:
Lots of changes, I don't get them all and the PR's text doesn't cover everything, as far as I can tell. But I guess there's not much to improve as all the other devs are missing, so…

**changes-approved**

adiroiban:
not that many changes. beside the alpine one.. there is a change to use hexdigest istead of binary digest and to detect the version on AIX... which were "backports" from server.

But yes, they are dev level changes.

Thanks for the review!

PR #78, opened 2017-09-26
[#4353] Run cleanup in reverse order as they are added.

adiroiban:
Scope
=====

This fix the cleanup code which should call the cleanup in reverse order.

Changes
=======

Run in reverse.

As a drive by, I have start moving away from python-future as it is bad :) and just use unicode_literals

reviewers: @lgheorghiu 

Not much to test. Here for compliance as there is no one else to review


adiroiban:
needs-review

PR #79, opened 2017-10-26
[#4427] Full tests on Alpine.

adiroiban:
Scope
=====

This enables all tests on Alpine.

Previously the PAM tests was skipped due to seg fault.

It also does some cleanup to fix the failure of server tests on Windows.

Changes
=======

Use the latest alpine version.

I am experimenting with flaky test handling, first in compat and if looks good we can push it to server.
There is a test which is tried 3 times.

While working on this, I have converted from future to six.
Future is ugly as it creates new classes, while six is just a thin layer and always uses stdlib classes.

As a drive-by change, I have fixed a twisted test case error which might happen with threaded reactor on clean.

While checking the coverage I found some code which was not used.

reviewers: @dumol @bgola 

I am using this PR to check that all testa are passing on compat, as it took some time since we made changes to it.

For the near future, I am thinking of setting up daily test run on master to make sure that tests still pass on all the slaves.

check that changes are ok.

---------

@dumol  I hope that I am using the right alpine version.

on openbsd i have executed

```
$ for gid in `getent group | grep str | cut -d: -f1`; do sudo groupdel $gid; done
```

to clen up some groups which were left after tests which did not manage to have a proper cleanup.

-----------

@bgola check that conversion from future to six is ok. The diff coverage is not 100% but the change should increase the coverage.

Thanks!


adiroiban:
needs-review

adiroiban:
needs-review

adiroiban:
needs-review

dumol:
Glad to see the PAM binaries built specifically for Alpine are working. 

The git rev for Alpine is good, as updated here also: https://github.com/chevah/server/tree/4332-alpine-linux. Don't you also need to updated `paver.sh` to get Alpine detection? The latest version in master should be good.

Not much to say otherwise, so…

**changes-approved**

PR #80, opened 2017-11-01
[#4387] Use coverator for coverage publishing.

bgola:
Scope
=====

Use a new tool instead of codecov to publish coverage data.

Changes
=====

I've added the `[paths]` section to the `.coveragerc` configuration file. This is needed for better path handling when combining different coverage data files on a different machine/path where the data was gathered.

Also implemented a new publish_coverage that uses the chevah-coverage tool instead of codecov to submit the coverage data.

The new publish_coverage uses some git info that are not available in the GitBrink API so there is some low level git usage, but this should be gone when we move this to brink.

reviewers: @adiroiban 

Please check if the changes make sense.

Running `paver test_coverage && paver coverage_publish` should publish the data to `http://coverage.chevah.com:8080`.

The server will only generate the HTML reports after 7 different builders publish reports, to simulate this manually you can run:

`BUILDER_NAME=some-name paver coverage_publish`

I'm running a test instance of the coverage server in `http://172.20.245.1:8080` so for the command above to work you need to modify your your hosts file.


bgola:
@adiroiban  this is the start of at least 3 PRs for this whole work regarding the coverage tool. I am not sure if this is the best way to divide it or if I should create different tickets for the PRs on trac also.

adiroiban:
each task should have separate ticket.

This looks like you want to enable the new coverage reporting tool to compat .

bgola:
ok, i've changed the ticket number to a new ticket for this specific task

bgola:
needs-review

adiroiban:
the PR title and branch name are not in sync.

I was also expecting to see the coverage sent to the new tool and check that the path configuration works.

You can implement a custom coverage_publish in pavement.py just for testing... and if all is ok you can update brink

needs-changes

bgola:
the coverage_publish i have implemented in brink because i needed to implement new functionalities to the BrinkGit.

I needed to create this branch so i can at least test a real case of cloning from github and combining on my vbox. without this .coveragerc file this would never work.

but now i have tested and we can maybe ignore this PR for now.

 i am still not sure regarding this brink X pavement situation. i can't implement my new coverage_publish because it uses BrinkGit.origin and I have also fixed BrinkGit.revision, it was returning the tree hash instead of the revision.

maybe I should first open a PR in brink and then, once approved, introduce the new brink on this PR?

or I can implement this origin and revision properties accessing the BrinkGit object from the pavement.py in compat.. just for now.

adiroiban:
> maybe I should first open a PR in brink and then, once approved, introduce the new brink on this PR?

brink has some tests (just to be there) so yes, you can start in brink for the publish (client-side) part.

bgola:
i ended up introducing the coverage_publish here for simplicity. 
and updated the PR description. 

The coverage report via PQM won't work for now because the slaves are going to try to publish the report to coverage.chevah.com which is unknown to them, but it would be good if I can get some feedback from the manual run, and also some review for the work done on https://github.com/chevah/chevah-coverage

needs-review

thanks

adiroiban:
Looks good. Thanks.

We also need `codecov/project/tests` ...  but this can be done in a separate PR.

For an initial PR is ok.... looking forward having this code in brink... and in compat to just use brink.

----------

Instead the harcoded 7 reports, for a separate PR the server can check the list of required buildbot/* commit status. Once all are red or green, it can send the reports.

You can also set up a GitHub hooks server and GitHub will tell you when each builder is done.

-----------

Instead of the hardcoded URL, the whole code should read the value from SETUP. maybe at `SETUP['test']['coverate_url']`

once brink is updated and we have URL in SETUP this can be merged.

needs-changes

pavement.py
```diff
@@ -112,7 +110,7 @@
     'mock',
 
     'coverage==4.0.3',
-    'codecov==2.0.3',
+    'coverator==0.1',
```
adiroiban: is better to go with semver scheme... so 0.1.0

pavement.py
```diff
@@ -482,13 +481,21 @@ def test_ci(args):
         args = [args]
     test_type = env.get('TEST_TYPE', 'normal')
 
+    if test_type == 'py3' or test_type == 'os-independent':
+        skip_coverage = True
```
adiroiban: is this used?

pavement.py
```diff
@@ -112,7 +109,6 @@
     'mock',
 
     'coverage==4.0.3',
```
adiroiban: we also have coverage==4.4.1 ... for which we have wheels

bgola:
i think we are done with this one?

needs-review

bgola:
actually, needs-changes

the name of the PR is still wrong compared to the branch.

bgola:
needs-review

PR #82, opened 2017-11-08
[#4468] Fix getAttributes on Windows for broken links.

adiroiban:
Scope
=====

getAttributes will success on Windows when link is broken, but will fail on Unix.

This unifies the behaviour.

Changes
=======

For windows, check that path exists before getting attributes

reviewers: @bgola 

check that changes make sense


adiroiban:
needs-review

PR #83, opened 2017-12-19
[#4588] Add option to except thread names on ChevahTestCase tearDown.

bgola:
Scope
=====

This adds an option to ignore certain threads during the tearDown.

Changes
=======

Added a class attribute to ChevahTestCase that can be overridden by subclasses / other test cases.

reviewers: @adiroiban 

Please check if it makes sense. Thanks!

### DIFF
--- a/chevah/compat/testing/testcase.py
+++ b/chevah/compat/testing/testcase.py
@@ -40,7 +40,7 @@ from chevah.compat import (
     SuperAvatar,
     )
 from chevah.compat.administration import os_administration
-from chevah.compat.testing.assertion import AssertionMixin
+from chevah.compat.testing.assertion import AssertionMixin, Contains
 from chevah.compat.testing.mockup import mk
 from chevah.compat.testing.constants import (
     TEST_NAME_MARKER,
@@ -792,10 +792,11 @@ class ChevahTestCase(TwistedTestCase, AssertionMixin):
     os_version = _get_os_version()
     cpu_type = _get_cpu_type()
 
+    # List of thread names to ignore during the tearDown.
     excepted_threads = [
         'MainThread',
         'threaded_reactor',
-        'PoolThread-twisted.internet.reactor',
+        Contains('PoolThread-twisted.internet.reactor'),
         ]
 
     # We assume that hostname does not change during test and this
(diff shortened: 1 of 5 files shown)


bgola:
needs-review

I'm not 100% satisfied with the automated tests but it is an improvement..

chevah/compat/tests/normal/test_testing.py
```diff
@@ -22,3 +26,53 @@ def test_avatar_unicode(self):
...
+        thread.start()
+
+        with self.assertRaises(AssertionError) as context:
+            self.tearDown()
+
+        self.assertContains('TestThread', context.exception.message)
+
+        # Stop the thread.
+        event.set()
+        time.sleep(0.1)
```
adiroiban: do we need the sleep? maybe do a thread.join() to wait for it to finish.

chevah/compat/tests/normal/test_testing.py
```diff
@@ -22,3 +26,53 @@ def test_avatar_unicode(self):
...
+        self.tearDown()
+
+        self.assertTrue(thread.is_alive())
+        event.set()
+
+        # Wait for the thread to stop.
+        time.sleep(0.1)
+        self.assertFalse(thread.is_alive())
+
+        self.excepted_threads = old_excepted_threads[:]
```
adiroiban: maybe use addCleanup to revert this... but I was thinking that the default testcase already has a helper which restore all instance variables after a test, so we don't need to restore them

chevah/compat/testing/testcase.py
```diff
@@ -792,6 +792,12 @@ class ChevahTestCase(TwistedTestCase, AssertionMixin):
     os_version = _get_os_version()
     cpu_type = _get_cpu_type()
 
+    excepted_threads = [
```
adiroiban: maybe we should add a comment to document the intended use of this list.

chevah/compat/testing/testcase.py
```diff
@@ -819,12 +825,11 @@ def tearDown(self):
...
-                if thread_name == 'MainThread':
-                    continue
-                if thread_name == 'threaded_reactor':
+                if thread_name in self.excepted_threads:
                     continue
-                if thread_name.startswith(
-                        'PoolThread-twisted.internet.reactor'):
+                if ('PoolThread-twisted.internet.reactor' in
+                        self.excepted_threads and thread_name.startswith(
+                        'PoolThread-twisted.internet.reactor')):
```
adiroiban: I think that we should allow Contain('PoolThread-twisted.internet.reactor') as part of the list in order to have a more relaxed matching.

bgola:
I did with Contains() by implementing __eq__ in Contains... what do you think?
needs-review
thanks

bgola:
the ticket id was wrong.

needs-review

bgola:
i updated https://github.com/chevah/server/pull/1342 to use this version / feature.

PR #85, opened 2018-03-08
[#4753] Always iterate the reactor with timeout.

adiroiban:
Scope
=====

This is an import for some changes which were done in chevah/server but which should have been here.

Changes
=======

Use a small timeout instead of None for reactor iteration.

As a drive by I moved more functions to compat and wrote tests, as in chevah/server they were just used, but without dedicated tests.

reviewers: @bgola 

check that changes make sense

thanks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 *.pyc
-build/
-build-*/
-dist/
+/build/
+/build-*
+/dist/
 *.egg-info/
-DEFAULT_VALUES
-cache/
+/DEFAULT_VALUES
+/cache
--- a/chevah/compat/testing/testcase.py
+++ b/chevah/compat/testing/testcase.py
@@ -498,10 +498,12 @@ class TwistedTestCase(TestCase):
 
         self._shutdownTestReactor()
 
-
     def executeDelayedCalls(self, timeout=None, debug=False):
         """
         Run the reactor until no more delayed calls are scheduled.
+
+        This will wait for delayed calls to be executed and will not stop
+        the reactor.
         """
         if timeout is None:
             timeout = self.DEFERRED_TIMEOUT
--- a/chevah/compat/testing/testcase.py
+++ b/chevah/compat/testing/testcase.py
@@ -302,7 +302,7 @@ class TwistedTestCase(TestCase):
             raise AssertionError(
                 'Reactor is not clean. %s: %s' % (location, reason))
 
-        if reactor._started:
+        if reactor._started:  # noqa:cover
             # Reactor was not stopped, so stop it before raising the error.
             self._shutdownTestReactor()
             raise AssertionError('Reactor was not stopped.')
--- a/chevah/compat/testing/testcase.py
+++ b/chevah/compat/testing/testcase.py
@@ -228,7 +228,7 @@ class TwistedTestCase(TestCase):
         Iterate the reactor.
         """
         reactor.runUntilCurrent()
-        if debug:
+        if debug:  # noqa:cover
             # When debug is enabled with iterate using a small delay in steps,
             # to have a much better debug output.
             # Otherwise the debug messages will flood the output.
@@ -258,7 +258,8 @@ class TwistedTestCase(TestCase):
             t = reactor.running and t2
             reactor.doIteration(t)
         else:
-            reactor.doIteration(False)
+            # To not slow down all the tests, we run with a very small value.
+            reactor.doIteration(0.000001)
 
     def _shutdownTestReactor(self, prevent_stop=False):
         """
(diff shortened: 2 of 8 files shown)


adiroiban:
win 2003 is failing due to TLS issue. @dumol I think that we can disable all win2003 slaves as I don't think that anybody is using it in production.

osx is failing due to pam

@bgola please give it a try

needs-review

dumol:
PAM issues in OS X / MacOS are probably caused by wrong binaries being deployed through `salt-master` after `storage` sub-module was removed. I am taking care of that with high prio, hopefully solved by the end of today.

dumol:
While at it I have:
 * Updated `paver.sh` from `python-package` master.
 * Improved `.gitignore` to specify root sub-dirs explicitly and to make possible the use of symbolic links for `cache` and `build-*` sub-dirs, as in `python-package`.

dumol:
Fixed `osx-108`/`macos-1012` slaves, they now have appropriate binaries. Not all PAM tests pass though, so I guess someone (@bgola?) still has to take a look to try making `compat` tests all green again. More at:
 * https://chevah.com/buildbot/builders/compat-osx-108/builds/7/steps/test/logs/stdio 
 * https://chevah.com/buildbot/builders/compat-macos-1012/builds/7/steps/test/logs/stdio

Have also updated Git version on the older `win-2003` build slave. I would not put off Windows 2003 as long as we have a Windows XP slave… `win-2003` is green:
 * https://chevah.com/buildbot/builders/compat-win-2003/builds/4

adiroiban:
@dumol I think that the tests are OK, is just that the PAM module is broken... now it accepts everything, even wrong username or password.

The strange thing is that we have similar tests in chevah/server for PAM and they passed in the past... not they are failing due to latest change

dumol:
I see most PAM-related server tests were disabled some time ago on OS X. Here's a green build from the end of February: https://chevah.com/buildbot/builders/server-osx-108/builds/171.

It has 3 / 4 PAM-related tests disabled, namely:

`
chevah.server.tests.elevated.normal.authentication.test_os:TestOSAuthentication.test_authenticate_pam_disabled_valid_credentials
chevah.server.tests.elevated.normal.authentication.test_os:TestOSAuthentication.test_authenticate_pam_exclusive_valid_credentials
chevah.server.tests.elevated.normal.authentication.test_os:TestOSAuthentication.test_authenticate_pam_disabled_os_credentials
`

So lately the only change is for this test, which worked previously (but I bet it wasn't indicative of a working PAM setup on OS X):

`
chevah.server.tests.elevated.normal.authentication.test_os:TestOSAuthentication.test_authenticate_pam_exclusive_bad_pass
`

adiroiban:
> I see most PAM-related server tests were disabled some time ago on OS X. Here's a green build from the end of February: https://chevah.com/buildbot/builders/server-osx-108/builds/171.

Thanks a lot  @dumol for the investigation.

Yes. The test were auto skipped as there is no dedicated pam service (/etc/pam.d/chevah-pam-test)
If the slave has no /etc/pam.d/chevah-pam-test file, the tests are skipped.

I guess that at that time the pam was not there, and now we have a /etc/pam.d/chevah-pam-test.

I suggest to remove the etc file for now and try to work on a fix at https://github.com/chevah/salt-master/pull/363

@dumol you can assign 4770 to me and I will take a look at making the PAM work on OSX.

Thanks again!

adiroiban:
@dumol I see that obsd is failing... I think that is due to poor coding in creating the testing accounts.

@dumol are you still using OBSD when working for ProAtria?
If not, I would like to remove support for OBSD and I don't have time to keep it active. Sorry

Thanks

dumol:
Files `/etc/pam.d/chevah-pam-test` on OS X / MacOS slaves are dated December 21, so the files were present at the time of that `server` build, at the end of February. My understanding (from `compat` code) is that all PAM-related tests are skipped when this file is missing. So I guess it was a more complex issue on `server` as one of the tests was actually executed?

Also, even for latest `server` tests I see those three PAM-related tests are skipped, eg. https://chevah.com/buildbot/builders/server-osx-108/builds/186/steps/test/logs/stdio.

In regards to the OpenBSD-related code, it's not necessary to have all things working to use OpenBSD from my side. I'm fine if I have a working `python-package` and `test_remote` works. And I'll take care of `python-package`, even if that means manually uploading latest tarballs when `compat` tests fail. Thanks!

adiroiban:
>  So I guess it was a more complex issue on server as one of the tests was actually executed?

Right now is hard to tell why they were skipped. I will update the tests to make it clear why the tests were skipped.

> And I'll take care of python-package, even if that means manually uploading latest tarballs when compat tests fail. Thanks!

I can disable the test which needs root for OBSD... The code works, but is the testing infrastructure which fails... in this way compat will still be green and no need for manual upload.

Will do that in another ticket. Thanks!

dumol:
OK, fine. Thanks for the OpenBSD support in `python-package`.

I couldn't help taking a look at the PAM-related tests in `server` too… I see `test_authenticate_pam_disabled_os_credentials` and `test_authenticate_pam_exclusive_valid_credentials` in `server` were recently specifically disabled for OS X, associated ticket added: https://trac.chevah.com/ticket/4606. 

The other one of the three disabled PAM-related tests seems to be skipped on everything but Solaris (where the `if` clause is not reached because of the previous condition): https://github.com/chevah/server/blob/a8887e4efe9c53a399b15a9c9a85a059c7da4853/chevah/server/tests/elevated/normal/authentication/test_os.py#L228. I could issue a ticket for this, and even fix it if there are no implications going much deeper…

The only remaining mystery for the `server` tests is how `test_authenticate_pam_exclusive_bad_pass` passed with the Debian binaries on OS X / MacOS. :-]

PR #86, opened 2018-04-15
[#4874] Raise OSError on setting permission instead of AttributeError.

adiroiban:
Scope
=====

This updated the Windows behaviour to look a bit closer to Unix when setting owner and group.

Changes
=======

Always raise OSError

As a drive by I created separeate tests for code which has different behaviour on Linux and Windows.

As a drive-by change I wrote a test for setAttribute for time uid and mode.

reviewers: @dumol 

@dumol  Check to see if you can make sense of the code.

I would also like to use this PR to make sure that compat test are green in order to support the python-package.

I am also updating the openbsd part.


adiroiban:
needs-review

chevah/compat/tests/normal/test_filesystem.py
```diff
@@ -101,15 +101,13 @@ def test_getPath_upper_paths(self):
         self.assertEqual(u'/c', path)
 
 
-class TestLocalFilesystem(CompatTestCase, FilesystemTestMixin):
+class DefaulFilesystemTestCase(CompatTestCase, FilesystemTestMixin):
```
dumol: Curious why this is not `DefaultFilesystemTestCase`…

dumol:
**changes-approved**

chevah/compat/tests/normal/test_filesystem.py
adiroiban: typo :) Will Fix. Thanks!

PR #87, opened 2018-04-30
[bad-id] Add support for UNC and links to Windows Shares.

adiroiban:
see https://github.com/chevah/compat/pull/88

### DIFF
--- a/chevah/compat/interfaces.py
+++ b/chevah/compat/interfaces.py
@@ -427,6 +427,8 @@ class ILocalFilesystem(Interface):
         """
         Return True if segments points to an existing path.
 
+        It will return True for broken links.
+
         This should check that a file exists without reading the file or folder
         in order to not trigger errors due to permissions errors.
         """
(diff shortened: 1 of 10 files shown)


adiroiban:
coverator still not working, but the other slaves are green.

will look at coverator later, but that is a separte issue

needs-review

hcs0:
Looks like it is the wrong ticket ID - could be https://trac.chevah.com/ticket/3543

adiroiban:
> Looks like it is the wrong ticket ID - could be https://trac.chevah.com/ticket/3543

3543 is about allowing direct access to the UNC, not via symbolic links.

This ticket is about creating a symbolic link to an UNC and using that simbolic link.

The symbolic link has normal local path.

So for now, you can access the files from \\server\share-name via c:\users\some-link 

3543 is about being able to access the \\server\share-name directly without the need for a symbolic link

hcs0:
@adiroiban 4861 is writing a user guide (recursive vs non recursive) https://trac.chevah.com/ticket/4861

PR #90, opened 2018-06-14
[#4969] Add support for virtual folders.

adiroiban:
Scope
=====

This adds support for creating virtual folders (virtual segments) on top of the normal available segments.

This will work like a symlink and is available for both locked and unlocked filesystems.

Changes
=======

Add support for virtual folders and error handling when trying to mutate a virtual folder.

reviewers: @dumol 

The filesystem handling is part of chevah/compat so we have this change here... but the main tests will be done in a server dedicated PR

This PR is big as there are a lot of tests.

### DIFF
--- a/chevah/compat/avatar.py
+++ b/chevah/compat/avatar.py
@@ -22,13 +22,15 @@ class FilesystemAvatar(HasImpersonatedAvatar):
 
     def __init__(
         self, name, home_folder_path, root_folder_path=None,
-        lock_in_home_folder=True, token=None
+        lock_in_home_folder=True, token=None,
+        virtual_folders=(),
             ):
         self._name = name
         self._home_folder_path = home_folder_path
         self._root_folder_path = root_folder_path
         self._token = token
         self._lock_in_home_folder = lock_in_home_folder
+        self._virtual_folders = virtual_folders
 
         if not isinstance(self._home_folder_path, text_type):
             raise RuntimeError('home_folder_path should be text.')
@@ -72,6 +74,13 @@ class FilesystemAvatar(HasImpersonatedAvatar):
         '''Return avatar's name.'''
         return self._name
 
+    @property
+    def virtual_folders(self):
+        """
+        See: :class:`IFileSystemAvatar`
+        """
+        return self._virtual_folders
+
 
 class FilesystemOSAvatar(FilesystemAvatar):
     """
@@ -90,7 +99,7 @@ class FilesystemOSAvatar(FilesystemAvatar):
 
 class FilesystemApplicationAvatar(FilesystemAvatar):
     """
-    Application avatar interacting with thefilesystem.
+    Application avatar interacting with the filesystem.
     """
 
     @property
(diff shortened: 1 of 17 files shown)


adiroiban:
So.... there was a big bug/error in compat and a lot of tests were skipped.

Now we run more tests.

* HP-UX runs out of memory with this PR...I can try the new python-package version to see the memory usage
* Hetiur is down... and all slaves are on Hetiur... I prefer to keep the previous VM up and running for 1 month
* The Ubuntu 18.04 is strange... it looks like it is setup to allow setting the group for a file to a group to which the user is not member... or this is what the tests shows https://chevah.com/buildbot/builders/compat-ubuntu-1804/builds/4/steps/test/logs/stdio

needs-review

dumol:
True, there's a problem with Hetiur, I've filled https://trac.chevah.com/ticket/4977 for it. 

We had VMs on Hetiur running alongside the original VM for about two weeks already. This issue has cropped up only during this week-end, but things are under control now that I know the cause.  It's easy to reactivate the old VMs if needed, but I would like to try to rely on Hetiur's VM exclusively from now on and delete the old ones in a month if things are fine.

adiroiban:
> In addition, it would help to document this somehow. There's a small note in the release notes, but outside of that it's not that clear what's with these virtual folders.

The end user documentation is found in chevah/server https://github.com/chevah/server/pull/1430

We don't have documentation for chevah/compat, other than the tests which should serve as documentation and examples of how you can use chevah/compat.

-------

Thanks for the review.

This PR will still be opened while the chevah/server PR is tested, as if we found issues there, they might need a fix here.

--------------

I have checked the test_iterateFolderContent_big 
While it takes a long time it is there also to check the limits of the OS.
For ex hpux, freebsd , openbsd don't allow more than 32765 members in a folder
And AIX is just super slow.
But for all other systems we test with 45.000 members in a folder.

adiroiban:
Thanks for the review.

> Also, beware that I tried to use varied filesystems where possible,

+1 Is very good to have as many file system as possible.

------------

Since we decided that we want to raise an error on path shadowing (virtual path with the same name as real path) and not to support path overlay (virtual path which is located in the same folder with other real folders) I will rewrite part of this.

Also, I forgot to add handling for case-insensitive for Windows paths.

needs-changes

adiroiban:
hpux memory issue is due to Coverage report and I this is a know error and I have disabled the coverage report

For Sparc and ARM64 I have reduced the number of files created... so that the tests are faster.

Hope that now all tests are green.

-------

@dumol we need a script on the slaves to do weekly cleanup of temp folders and accounts which are left after some failed tests.

In an ideal world we would have tests which even when they fail, they can do the cleanup, but this is not always possible.

For example, if a test takes to long and buildslaves kills the test, the cleanup is no longer executed.

Please create a ticket and let me know if you need help identifying what need to be cleaned.

Thanks!

adiroiban:
needs-review

dumol:
Happy to see all build slave turning green with latest changes.

Have added https://trac.chevah.com/ticket/4983 for removing leftovers from unclean tests.

Beware that both NTFS and HFS+ have support for case sensitivity, although it's rare to have it used, especially in Windows where I think you have to eschew Win32 to use it. And it looks like APFS continues to offer both case-sensitive and case-insensitive options in macOS, with the latter still being the default.

**changes-approved**

PR #91, opened 2018-09-19
[#5072] Add filename to OSError.

adiroiban:
Scope
=====

This updates the compat code to add filename when OSError has no filename.
For example on Windows.

Changes
=======

Refactor the code to add filename.

As a drive by, update the dev env for latest buildbot with local passwords.

reviewers: @dumol 

We can use this PR to update the python package.

Please check what Python-package should I use.


adiroiban:
needs-review

dumol:
Python package version is good, the last one with PyCrypto, should work on all supported platforms, including HP-UX.

Glad to see tests pass.

**changes-approved**

PR #92, opened 2018-10-03
[#5082] Return attributes from folder iterations.

adiroiban:
Scope
=====

This updates the folder iteraction fuction to return file attributes, instead of just names.
In this way, you don't need an extra call to get the attributes for each member.

Changes
=======

Convert scandir attributes to FileAttributes, update test.

reviewers: @dumol 

Not much to review. Hope all slaves are ok and we can add arm64 works.

### DIFF
--- a/chevah/compat/posix_filesystem.py
+++ b/chevah/compat/posix_filesystem.py
@@ -632,10 +632,13 @@ class PosixFilesystemBase(object):
 
             child_segments = virtual_segments[segments_length:]
 
-            result.append(self._getPlaceholderAttributes(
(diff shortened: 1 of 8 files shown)


adiroiban:
needs-review

dumol:
Fine by me! Used the opportunity to update `paver.sh` which in return required Debian packages with PyCrypto. As those will be needed in `server` too, I went ahead and built them in https://github.com/chevah/python-package/pull/118 and updated `paver.conf` to use them.

**changes-approved**

PR #93, opened 2019-02-07
[#5200] Overwrite destination on rename in Windows

adiroiban:
Scope
=====

On  Windows, default rename fails if destination exists.

But we don't want to fail so that we have the same behaviour on all OS.

Changes
=======

Handle error on rename on Windows and delete the destination

As a drive-by I have also updated the paver.sh and paver.conf

reviewers: @dumol 

This is more FYI.

### DIFF
--- a/chevah/compat/nt_filesystem.py
+++ b/chevah/compat/nt_filesystem.py
@@ -703,10 +703,23 @@ class NTFilesystem(PosixFilesystemBase):
             raise error
 
     def rename(self, from_segments, to_segments):
-        '''See `ILocalFilesystem`.'''
+        """
+        See `ILocalFilesystem`.
+        """
         with self._windowsToOSError(from_segments):
-            return super(NTFilesystem, self).rename(
-                from_segments, to_segments)
+            try:
+                return super(NTFilesystem, self).rename(
+                    from_segments, to_segments)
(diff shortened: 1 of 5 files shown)


adiroiban:
@dumol this is more of a review to check the current state of compat buildslaves :)

There was not much activity here lately... as we move to supporting only Windows / Linux ..

needs-review

dumol:
I see the current list of required build targets is a bit different than what you said on IRC last week. I'm fine with whatever, have updated Zabbix monitoring accordingly, so there should be at least one slave up for every required OS at all times.

**changes-approved**

adiroiban:
The list is  different as obsd / aix-53 was green on compat with zero effort...as long as they work, we can keep them.

But, if keeping OpenBSD/AIX-53 slave alive is extra work for you, feel free to stop them and I will update the compat list.

Thanks!

dumol:
Thanks! OpenBSD I kind of need, to have a Python package for myself. But it's been care free for some years now, I only do OS upgrades on it before upgrading on my desktop.

An AIX 5.3 slave runs as vWPAR on Eliseni, not much work for that either. I recently checked on Nades, the stand-alone AIX 5.3 installation, that is also fully functional, but shut down, to save some electricity.

PR #94, opened 2019-03-14
[#5226] Minor test fixes related to Windows 2019.

dumol:
Scope
=====

`TestSystemUsers.test_getHomeFolder_posix` fails for home dirs with ending slash. This is because in salt-master's pillars I've updated buildslave's home dir to include the ending (forward or backward) slash in order to minimise the number of if clauses needed for Windows support in #5185. Eg. https://chevah.com/buildbot/builders/python-package-rhel-7/builds/10/steps/test_1/logs/stdio.

Also, `TestLocalFilesystemLocked.test_getAttributes_link_file_outside` fails on Windows 2019. Eg. https://chevah.com/buildbot/builders/python-package-win-2019/builds/10/steps/test_1/logs/stdio

Changes
=======

On POSIX systems, strip `/` from the end of a home dir when getting it.

On Windows, check for `c:\windows\system32\xcopy.exe` instead of `c:\windows\explorer.exe`, which is missing from Windows Server Core.

Also adapted a couple of elevated Windows tests for Windows 2019.

reviewers: @adiroiban 

Please double-check my changes.

Run the tests.

### DIFF
--- a/chevah/compat/nt_filesystem.py
+++ b/chevah/compat/nt_filesystem.py
@@ -716,7 +716,7 @@ class NTFilesystem(PosixFilesystemBase):
                 if error.errno != errno.EEXIST:
                     # Not a file already exists error.
                     raise error
-                # Try to remove the fine, and then rename one more time.
+                # Try to remove the file, and then rename one more time.
                 self.deleteFile(to_segments)
                 return super(NTFilesystem, self).rename(
                     from_segments, to_segments)
--- a/chevah/compat/nt_filesystem.py
+++ b/chevah/compat/nt_filesystem.py
@@ -772,7 +772,7 @@ class NTFilesystem(PosixFilesystemBase):
                     path,
                     win32security.SE_FILE_OBJECT,
                     win32security.DACL_SECURITY_INFORMATION,
-                    None,
+                    user_sid,
                     None,
                     d_acl,
                     None,
--- a/chevah/compat/testing/testcase.py
+++ b/chevah/compat/testing/testcase.py
@@ -805,10 +805,9 @@ def _get_os_version():
     """
     On non-Linux this is just the os_name.
 
-    On Linux is the distribution name, without the version.
+    On Linux is the distribution name and the version.
 
-    On Windows it is the `nt` followed by the version, which
-    consists of major version, minor version, build number.
+    On Windows it is the `nt` followed by the major and minor NT version.
     It is not the marketing name.
     We only support the Windows NT family.
     See: https://en.wikipedia.org/wiki/Windows_NT#Releases
@@ -818,8 +817,8 @@ def _get_os_version():
     See: https://en.wikipedia.org/wiki/MacOS#Release_history
     """
     if os.name == 'nt':
-        parts = platform.win32_ver()[1].split('.')
-        return 'nt-%s.%s.%s' % (parts[0], parts[1], parts[2])
+        parts = platform.version().split('.')
+        return 'nt-%s.%s' % (parts[0], parts[1])
 
     # We are now in Unix zone.
     os_name = os.uname()[0].lower()
--- a/chevah/compat/testing/testcase.py
+++ b/chevah/compat/testing/testcase.py
@@ -133,33 +133,37 @@ class TwistedTestCase(TestCase):
             result.append(text_type(delayed.func))
         return '\n'.join(result)
 
-    def _threadPoolQueueSize(self):
+    def _threadPoolQueue(self):
         """
-        Return current size of thread Pool, or None when threadpool does not
+        Return current tasks of thread Pool, or [] when threadpool does not
         exists.
         """
         if not reactor.threadpool:
-            return 0
-        else:
-            return reactor.threadpool.q.qsize()
+            return []
+
+        # Consume all the
+        result = []
+        while not reactor.threadpool.q.empty():
+            result.append(reactor.threadpool.q.get()[1])
+        return result
 
     def _threadPoolThreads(self):
         """
-        Return current threads from pool, or None when threadpool does not
-        exists.
+        Return current threads from pool, or empty list when threadpool does
+        not exists.
         """
         if not reactor.threadpool:
-            return 0
+            return []
         else:
             return reactor.threadpool.threads
 
     def _threadPoolWorking(self):
         """
-        Return working thread from pool, or None when threadpool does not
-        exists.
+        Return working thread from pool, or empty when threadpool does not
+        exists or has no job.
         """
         if not reactor.threadpool:
-            return 0
+            return []
         else:
             return reactor.threadpool.working
 
@@ -237,7 +241,7 @@ class TwistedTestCase(TestCase):
                 u'threads: %s\n'
                 u'writers: %s\n'
                 u'readers: %s\n'
-                u'threadpool size: %s\n'
+                u'threadpool queue: %s\n'
                 u'threadpool threads: %s\n'
                 u'threadpool working: %s\n'
                 u'\n' % (
@@ -245,7 +249,7 @@ class TwistedTestCase(TestCase):
                     reactor.threadCallQueue,
                     reactor.getWriters(),
                     reactor.getReaders(),
-                    self._threadPoolQueueSize(),
+                    self._threadPoolQueue(),
                     self._threadPoolThreads(),
                     self._threadPoolWorking(),
                     )
@@ -311,13 +315,14 @@ class TwistedTestCase(TestCase):
         if len(reactor.threadCallQueue) > 0:
             raise_failure('threads', reactor.threadCallQueue)
 
-        if self._threadPoolQueueSize() > 0:
-            raise_failure('threadpoool queue', self._threadPoolQueueSize())
+        pool_queue = self._threadPoolQueue()
+        if pool_queue:
+            raise_failure('threadpoool queue', pool_queue)
 
-        if self._threadPoolWorking() > 0:
+        if self._threadPoolWorking():
             raise_failure('threadpoool working', self._threadPoolWorking())
 
-        if self._threadPoolThreads() > 0:
+        if self._threadPoolThreads():
             raise_failure('threadpoool threads', self._threadPoolThreads())
 
         if len(reactor.getWriters()) > 0:  # noqa:cover
@@ -880,6 +885,9 @@ class ChevahTestCase(TwistedTestCase, AssertionMixin):
     excepted_threads = [
         'MainThread',
         'threaded_reactor',
+        'GlobalPool-WorkerHandler',
+        'GlobalPool-TaskHandler',
+        'GlobalPool-ResultHandler',
         Contains('PoolThread-twisted.internet.reactor'),
         ]
 
--- a/chevah/compat/testing/testcase.py
+++ b/chevah/compat/testing/testcase.py
@@ -807,18 +807,19 @@ def _get_os_version():
 
     On Linux is the distribution name, without the version.
 
-    On Windows it is the `nt` followed by the major and minor NT version.
-    IS not the marketing name.
-    We only support Windows NT family.
+    On Windows it is the `nt` followed by the version, which
+    consists of major version, minor version, build number.
+    It is not the marketing name.
+    We only support the Windows NT family.
     See: https://en.wikipedia.org/wiki/Windows_NT#Releases
 
     On OSX it returns `osx` followed by the version.
-    Is not the Darwin version.
+    It is not the version of the underlying Darwin OS.
     See: https://en.wikipedia.org/wiki/MacOS#Release_history
     """
     if os.name == 'nt':
-        parts = platform.version().split('.')
-        return 'nt-%s.%s' % (parts[0], parts[1])
+        parts = platform.win32_ver()[1].split('.')
+        return 'nt-%s.%s.%s' % (parts[0], parts[1], parts[2])
 
     # We are now in Unix zone.
     os_name = os.uname()[0].lower()
--- a/chevah/compat/tests/elevated/test_capabilities.py
+++ b/chevah/compat/tests/elevated/test_capabilities.py
@@ -77,8 +77,8 @@ class TestProcessCapabilities(FileSystemTestCase):
                 'SeIncreaseWorkingSetPrivilege:0, SeTimeZonePrivilege:0, '
                 'SeCreateSymbolicLinkPrivilege:0'
                 )
-            if self.os_version == 'nt-10.0':
-                # On Win 2016 we have an extra capability by default.
+            if self.os_version.startswith('nt-10.0'):
+                # Starting with Win 2016 there's an extra default capability.
                 expected_capabilities += (
                     ', SeDelegateSessionUserImpersonatePrivilege:0'
                     )
--- a/chevah/compat/tests/elevated/test_capabilities.py
+++ b/chevah/compat/tests/elevated/test_capabilities.py
@@ -77,7 +77,7 @@ class TestProcessCapabilities(FileSystemTestCase):
                 'SeIncreaseWorkingSetPrivilege:0, SeTimeZonePrivilege:0, '
                 'SeCreateSymbolicLinkPrivilege:0'
                 )
-            if self.os_version.startswith('nt-10.0'):
+            if int(self.os_version.split('.')[2]) >= 14393:
                 # Starting with Win 2016 there's an extra default capability.
                 expected_capabilities += (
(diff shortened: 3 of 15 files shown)


dumol:
I see that in `compat` some more tests fail on Windows 2019. From https://chevah.com/buildbot/builders/compat-win-2019/builds/3:
```
chevah.compat.tests.elevated.test_capabilities:TestProcessCapabilities.test_getCurrentPrivilegesDescription
chevah.compat.tests.elevated.test_filesystem:TestPosixFilesystem.test_getOwner
```

I'm not sure why they don't fail when `compat` tests are run for `python-package` and I'm reluctant to try to fix them as long as `server` tests don't have an issue with the current setup, eg. https://chevah.com/buildbot/builders/server-win-2019/builds/5/steps/test/logs/stdio

Note that currently there are no special privileges awarded to `buildslave` user on Windows 2019. It's just a regular member of the `Administrators` group.

**needs-review**

chevah/compat/tests/normal/test_system_users.py
```diff
@@ -40,7 +40,7 @@ def test_getHomeFolder_posix(self):
         Check getHomeFolder on Linux and Unix.
         """
         home_folder = system_users.getHomeFolder(
-            username=mk.username)
+            username=mk.username).rstrip('/')
```
adiroiban: We should not fix the test, but rather update the implementation so that getHomeFolder will always return the path without the trailing slash.

I will make the change now. Thanks!

adiroiban:
> chevah.compat.tests.elevated.test_capabilities:TestProcessCapabilities.test_getCurrentPrivilegesDescription

The buildslave account is expected to have SERESTORE and SEBACKUP privileges... in this way it should have the permissions to create Profiles (home folders) for other users.

So the buildslave account should have Backup/Restore permissions and also debug permissions.

Here are some command line info ... https://www.techrepublic.com/article/set-user-rights-using-the-ntrights-utility/

But maybe just adding to "Backup operator" group should be enough.

----------

> chevah.compat.tests.elevated.test_filesystem:TestPosixFilesystem.test_getOwner
This checks the Owner of the C:\Users\buildslave folder... it looks like on the other slaves the folder is owned by "Administrators" group.

Not sure how to fix it....  If by default the folder is owned by "buildslave" maybe we should update all the other slaves to match the permissions we observe in 2019...

Or change 2019 to have "Administrators" as the owner for the home folder.

----------


You don't observe these tests as we run the tests in 2 stages.
First non-root/non-admin tests.
If they pass, then we will also run the root/admin tests which are the "elevated" tests

chevah/compat/tests/normal/test_system_users.py
dumol: I beg to differ, but it's up to you how to fix it… My problem is your change has brought the issue back, see for example https://chevah.com/buildbot/builders/compat-ubuntu-1604/builds/5

dumol:
> > chevah.compat.tests.elevated.test_capabilities:TestProcessCapabilities.test_getCurrentPrivilegesDescription
> 
> The buildslave account is expected to have SERESTORE and SEBACKUP privileges... in this way it should have the permissions to create Profiles (home folders) for other users.
> 
> So the buildslave account should have Backup/Restore permissions and also debug permissions.
> 
> Here are some command line info ... https://www.techrepublic.com/article/set-user-rights-using-the-ntrights-utility/

Unfortunately this command is not available under Windows 2019 Server Core.


> But maybe just adding to "Backup operator" group should be enough.

I've tried that, but there's no change in the test output. Maybe it's not enough, or maybe the test looks for something else? See for https://chevah.com/buildbot/builders/compat-win-2019/builds/9/steps/test/logs/stdio for details and note that only on bs1i-win-2019 we have `buildslave` as member of the 'Backup operators' group, as this required rebuilding the build slave setup from scratch. 


> > chevah.compat.tests.elevated.test_filesystem:TestPosixFilesystem.test_getOwner
> > This checks the Owner of the C:\Users\buildslave folder... it looks like on the other slaves the folder is owned by "Administrators" group.
> 
> Not sure how to fix it.... If by default the folder is owned by "buildslave" maybe we should update all the other slaves to match the permissions we observe in 2019...
> 
> Or change 2019 to have "Administrators" as the owner for the home folder.

I think the last option is something that can only be done manually. I would rather fix the other slaves to use the automated Salt setup. I've added a Trac ticket for that and started working on it: https://trac.chevah.com/ticket/5233

In the mean time, can you add a FIXME in the code and accept both variants?


> You don't observe these tests as we run the tests in 2 stages.
> First non-root/non-admin tests.
> If they pass, then we will also run the root/admin tests which are the "elevated" tests

Thanks, this explains it.

dumol:
> Note that after changing the group for "buildslave" you need to logout and login again.
> As otherwise the buildslave session might run with the old permissions...similar to Linux.

I reapplied `services.buildslave` from scratch, so I have also removed user/group/files and rebooted. Now I see it wasn't really needed, I was fooled by my mistake of applying the state to all `bs1-win*` slaves, instead of just `bs1-win-2019*`. Which yielded some unexpected and cryptic errors.

Could be that I broke the first slaves for the other Windows version with this. But just slightly, only the default shell seems to be changed, so this should not be a problem until the first restart. In the mean time, I'll try to get the `services.buildslave` state to cleanly apply on all of them in https://trac.chevah.com/ticket/5233.


> For home folder owne, I prefer to have "Administrators" group as the owner for the buildslave home folder.
> This non-default value means that our code works at some extend, and is not just returning a default value.

Ok, will look into it for the `win-2019` slave through Salt in the same ticket.

dumol:
I don't understand the issues with the privileges. All required privileges are reported as enabled for user ``buildslave` on the ``win-2019`` slaves:
```
Privilege Name                            Description                                                        State
========================================= ================================================================== =======
SeIncreaseQuotaPrivilege                  Adjust memory quotas for a process                                 Enabled
SeSecurityPrivilege                       Manage auditing and security log                                   Enabled
SeTakeOwnershipPrivilege                  Take ownership of files or other objects                           Enabled
SeLoadDriverPrivilege                     Load and unload device drivers                                     Enabled
SeSystemProfilePrivilege                  Profile system performance                                         Enabled
SeSystemtimePrivilege                     Change the system time                                             Enabled
SeProfileSingleProcessPrivilege           Profile single process                                             Enabled
SeIncreaseBasePriorityPrivilege           Increase scheduling priority                                       Enabled
SeCreatePagefilePrivilege                 Create a pagefile                                                  Enabled
SeBackupPrivilege                         Back up files and directories                                      Enabled
SeRestorePrivilege                        Restore files and directories                                      Enabled
SeShutdownPrivilege                       Shut down the system                                               Enabled
SeDebugPrivilege                          Debug programs                                                     Enabled
SeSystemEnvironmentPrivilege              Modify firmware environment values                                 Enabled
SeChangeNotifyPrivilege                   Bypass traverse checking                                           Enabled
SeRemoteShutdownPrivilege                 Force shutdown from a remote system                                Enabled
SeUndockPrivilege                         Remove computer from docking station                               Enabled
SeManageVolumePrivilege                   Perform volume maintenance tasks                                   Enabled
SeImpersonatePrivilege                    Impersonate a client after authentication                          Enabled
SeCreateGlobalPrivilege                   Create global objects                                              Enabled
SeIncreaseWorkingSetPrivilege             Increase a process working set                                     Enabled
SeTimeZonePrivilege                       Change the time zone                                               Enabled
SeCreateSymbolicLinkPrivilege             Create symbolic links                                              Enabled
SeDelegateSessionUserImpersonatePrivilege Obtain an impersonation token for another user in the same session Enabled
```

But for some reason, the following privileges are equal to 0 on the older windows slaves, but not on Windows 2019, where we get:
```
SeBackupPrivilege:2, SeRestorePrivilege:2, SeDebugPrivilege:2
```
They look the same with the user belonging or not belonging to the "Backup Operators" group.

dumol:
For reference, couldn't fix the other test either. So far have tried changing ownership for `c:\users\buildslave` to the following: `buildslave`, `BUILTIN\Administrators` and `NT AUTHORITY\SYSTEM`. The test returns the same:
```
AssertionError: u'Administrators' != u'buildslave'
- Administrators
+ buildslave
```

adiroiban:
The test_getOwner test is not checking "c:\users\buildslave" but rather a newly create home folder.

There is a new user created for these tests and we check the home folder for that user.

Now, I don't know why the home folder for this new users is owned by "buildslave"


I have run this manual test.
While running a ssh session as "builldslave" I have executed"

```
cmd
net user chevah-test some-password /ADD
```

Them, I have login using chevah-test and I can see that the new home folder is created as 

```
03/15/2019  04:22 PM    <DIR>          BUILTIN\Administrators chevah-test  
```

so, it looks like there is a bug in our code.

The ugly part is that Windows 2019 is just `nt-10.0` same as Windows 2016, so our compat code can't tell if it runs on 2019 or 2016.

I think that we can leave this for a bit later, as right now I don't have time to look into Windows 2019.

I will on how to fix this for Windows 2019 :(

chevah/compat/nt_filesystem.py
```diff
@@ -772,7 +772,7 @@ def _setOwner(self, path, owner):
                     path,
                     win32security.SE_FILE_OBJECT,
                     win32security.DACL_SECURITY_INFORMATION,
-                    None,
+                    user_sid,
```
adiroiban: there might be an error here... but the current failure is in getOwner and not in setOwner

dumol: True, this is just a leftover from some changes I've tried related to `setOwner`.  I thought it's best to leave this small change in because it clarifies what that field is used for and the needed variable ( `user_id` above) is already set and used. But no strong feelings, I can back this out if you think otherwise.

dumol: Happy that this was the only thing out-of-order with the above commit, which prevents the failures for `getOwner` in Windows 2019! :-]

dumol:
@adiroiban, I tried to update the elevated `compat` tests for Windows 2019 too. This was both painful and satisfying at the same time, so I guess it was masochistic of me… :-]

Please double-check and improve/approve, so I can get forward with https://github.com/chevah/python-package/pull/127.

Thanks!

chevah/compat/tests/elevated/test_capabilities.py
```diff
@@ -77,11 +77,30 @@ def test_getCurrentPrivilegesDescription(self):
...
                 'SeCreateSymbolicLinkPrivilege:0'
                 )
-            if self.os_version == 'nt-10.0':
-                # On Win 2016 we have an extra capability by default.
+            if int(self.os_version.split('.')[2]) >= 14393:
+                # Starting with Win 2016 there's an extra default capability.
                 expected_capabilities += (
                     ', SeDelegateSessionUserImpersonatePrivilege:0'
                     )
+            if int(self.os_version.split('.')[2]) >= 17763:
```
adiroiban: I think that we need a special os_build variable.... it doesn't make sense to  construct the os_version with the build number just to deconstruct it each time we need to get the build number.

the above condition should be just

```
if self.os_build >= 14393:
    # Do something
````

File: chevah/compat/testing/testcase.py [hunk 4]
```diff
@@ -807,18 +807,19 @@ def _get_os_version():
...
     On OSX it returns `osx` followed by the version.
-    Is not the Darwin version.
+    It is not the version of the underlying Darwin OS.
     See: https://en.wikipedia.org/wiki/MacOS#Release_history
     """
     if os.name == 'nt':
-        parts = platform.version().split('.')
-        return 'nt-%s.%s' % (parts[0], parts[1])
+        parts = platform.win32_ver()[1].split('.')
+        return 'nt-%s.%s.%s' % (parts[0], parts[1], parts[2])
```
adiroiban: This needs to be reverted, and have a separate `os_build_number` or just `os_build` member.

we have os_family, os_name, os_version which are all not comparable.
They were created to make tests easy to write and read and to document the code itself, without the need for extra comments.

The current usage as `int(self.os_version.split('.')[2]) >= 14393` is not easy to write or read.

What is `14393` ... if you always have to go to https://en.wikipedia.org/wiki/Windows_10_version_history and see that 14393 is in fact Windows 10 1607

I would lilke to see something like

`if self.os_build >= self.os_builds.WIN10_1607`

or


`if self.os_build >= self.WIN10_1607`


---------

The build is comparable, but as an int... so each time we will have to use the build number in a test, we will have to extract it from the os_version, convert in int and then compare.

Is better to have the os_build available right away and a few constants are reference points.

dumol: Thanks for the review! 

I agree this OS version with a build number is an ugly problem. And probably could be addressed more elegantly. 

But I feel you haven't addressed the core issue at hand here. And that is having identical `os_version` strings for different versions of Windows. For now, only Windows Server 2016 and 2019 are affected, but it's possible that in the near future all supported version of Windows will be in the same situation!

The only argument I heard against this is _maybe they are not different and we should treat them the same_. But the need for separate changes here show they are not the same. Microsoft doesn't think they are the same. We also test them separately, with dedicated slaves etc. This problem will creep up in more and more places.

So my stance is that we have to improve `os_version` to *always* differentiate between major OS versions. As shown in my meagre attempt, the required changes are not that invasive.

**P.S.** Now that I look at the code with a fresh eye, I can see how we got here. Because for Windows we chose to only look at the first two "parts" of `platform.version()`. This was based on an assumption that proved wrong. I think we should not try to be smarter than upstream Python here, therefore we should use the whole version string. That includes the build number for a reason.

adiroiban:
We are using the os_version for 2 purposed:

* Detect API level
* Detect build slave

We should use the os_version only to detect the API level.

Is ok if WIndows 10, Windows Server 2016 and WIndows Server 2019 have the same os_version, as they provide the same API level.

We should use `self.hostname` to detect the slave.

-------

> But I feel you haven't addressed the core issue at hand here. And that is having identical os_version strings for different versions of Window.

The solution proposed by you work for the moment... but once we update the Windows 2016 slaves, they will have the same build number with 2019.

https://en.wikipedia.org/wiki/Windows_Server_2016#Version_1803

dumol:
> > But I feel you haven't addressed the core issue at hand here. And that is having identical os_version strings for different versions of Window.
> 
> The solution proposed by you work for the moment... but once we update the Windows 2016 slaves, they will have the same build number with 2019.
> 
> https://en.wikipedia.org/wiki/Windows_Server_2016#Version_1803

Sorry, it seems you got this wrong. Windows Server version 1809 (Semi-Annual Channel) shares the same **version number 1809** with Windows Server 2019. This is a different issue (for the version number, not the build number) for a different Windows server version (Windows Server 1809 from the Semi-Annual Channel, not Windows Server 2016).

So, to recap:

  * Windows Server 2016 has **version** 1607 and **build number** 14393 (with or without updates).
  * Windows Server 2019 has **version** 1809 and **build number** 17763 (with or without updates).

This is from https://docs.microsoft.com/en-us/windows-server/get-started/windows-server-release-info. On a related note, we can't use version because it would only cover Windows Server 2016 or newer.

Counterintuitively, both **version** and **build number** are fixed and don't change with the updates! All the nitty-gritty details at https://techcommunity.microsoft.com/t5/Windows-Server-Insiders/Windows-Server-2019-version-info/td-p/234472

chevah/compat/tests/elevated/test_capabilities.py
```diff
@@ -77,13 +77,14 @@ def test_getCurrentPrivilegesDescription(self):
...
+            if self.os_version == 'nt-10.0':
+                # On latest Windows there is an extra capability by default.
                 expected_capabilities += (
                     ', SeDelegateSessionUserImpersonatePrivilege:0'
                     )
-            if int(self.os_version.split('.')[2]) >= 17763:
-                # This is Win 2019 or newer, some privileges are different:
+            if 'win-2019' in self.hostname:
+                # On the Windows 2019 slave some privileges are different,
+                # as UAC is disabled:
```
dumol: I'm not sure this is the reason… Why do you say _UAC is disabled_? I haven't disabled it manually or through Salt. This is how things stand by default in Windows 2019. 

The only thing I've tried was to include `buildslave` user in the `Backup Operators` group, but that doesn't changes privileges, as `buildslave` is part of the `Administrators` group.

adiroiban:
bs1a-win-2012-x64-34 needs to be salted... or something else, as it looks out of sync.

all good from my part. Many thanks for your help here. Much appreciated

needs-review

dumol:
Thanks for fixing the tests!

I don't know what's the matter with the first `win-2012` slave, it's salted and rebooted, but tests return other results. The only difference I see is the second `win-2012` slave seems to be a bit more up to date, so I'm installing updates on the first one right now, hoping for the best…

PR #95, opened 2019-05-12
[#5267] Fix str vs unicode testing. Better text generation.

adiroiban:
Scope
=====

This contains a bunch of small fixed found while working on chevah/server

Changes
=======

* Fix Twisted test case cleanup when a test fails
* Fix testing that when 2 values are equal, if they are string, they also have the same type.
* Update text generation to include uppercase letters.

reviewers: @dumol 

This is more like a FYI.

@dumol  I think that we should enable OpenBSD 3.6 for compat to match your system.

With this branch, I hope that we can have a green build on all slaves.

For my test, the new `set_download_commands` from `paver.sh` is too verbose :)
I don't need to be reminded each time, I run `paver.sh`... maybe only show the output for `paver deps` or when curl/wget is really used.  
For example `paver.sh remote` doesn't uses wget/curl


adiroiban:
I hope slaves will be up again and can run the tests

needs-review

dumol:
True, the output is too verbose when setting download tool, have fixed that in the last commit.

Also updated `paver.conf` for OpenBSD 6.5. 

Buildmaster now has `macos-1013` in compat list, `obsd` was already present. Haven't touched the repo's master branch settings, so `macos-1014` is still required. Feel free to keep it or remove it.

I believe `osx-1012` should definitely be removed though. It's not a supported platform any more.

**changes-approved**

adiroiban:
Thanks for the review and update.

We can keep buildbot/compat-macos-1014 for now.. I will move "Required" to 1013

I think that we should remove from buildmaster for compat, the following slaves
* buildbot/compat-sles-11-sm 
* buildbot/compat-osx-1012

dumol:
> I think that we should remove from buildmaster for compat, the following slaves
> 
>     * buildbot/compat-sles-11-sm
> 
>     * buildbot/compat-osx-1012

Done

PR #96, opened 2019-06-17
[#4145] Raise CompatError on errors raised by userExists.

adiroiban:
Scope
=====

This updates the unix and windows implementation of userExists to raise a CompatError  when it fails to check.

Changes
=======

As a drive by, I have renamed `constants.py` to `constant.py` ... as a convention to only use singular.

reviewers: @dumol 

This is just a FYI.

I have not written any automated tests as this is ugly

To check the behaviour I have done the following

```
# chmod 600 /etc/passwd
```

and then from a normal user I have executed

```
$ paver test normal.test_system_users
```
and observe that `test_userExists_found` fails.


adiroiban:
needs-review

dumol:
Test works on `ubuntu-1604-arm64` as well.

**changes-approved**

PR #98, opened 2020-04-30
[#3228] Strict new file perms. Non-world readable PID file.

adiroiban:
Scope
=====

This updates compat so that new files have non word readable/writable permission by default.

The previous default behaviour was to use 666 + UMASK.
But in some cases, there was no umaks, of the umask didn't had a good value.

Changes
=======

Add a mode argument when opening file operations which might create a new file.

As a cleanup, I have remove the UTF-8 magic handling.
I have also removed the openforUpdating as I don't know what is that and is not used.

reviewers: @dumol 

I will create a separate PR for chevah/server to test the actual daemon permissions.

For now, I have created a demo daemon script which can be used for manual testing the changes

I was testing it as

```
$ paver build
$ ./build-ubuntu2004-x64/bin/python example/unix-daemon.py 
$ cat /tmp/py-daemon-example.log
Starting in fork...
1588267432.14 Still alive.
1588267433.14 Still alive.
$ cat /tmp/py-daemon-example.pid 
560687
$ kill 560687
$ cat /tmp/py-daemon-example.pid 
cat: /tmp/py-daemon-example.pid: No such file or directory
```


adiroiban:
needs-review

adiroiban:
Thanks for the review. 

Where have you found the folder permissions issue?
Note that under normal circumstances, those values are applied together with the umask.

It just that somehow, the daemon / fork part is not observing the umask, hence we are left with this PID issue.

adiroiban:
but I think that we can also change the default value for directories.

needs-changes

dumol:
> Where have you found the folder permissions issue?

Judging by the second comment, I guess it's no longer a question, but for later reference: `chevah/compat/constants.py`.

adiroiban:
This is the fix for #213

PR #99, opened 2020-06-14
[#2042] Local copy no stats and Win i18n fix. Run on Linux from Unicode path. Run tests on Githu Actions.

adiroiban:
Scope
=====

This updates the local file copy to only copy data without attributes.
In this way, we no longer have 2 revisions (on versioned filesystems) for file on copy.

It also fixes the i18n tests.

It is a fix for running from Unicode path on Linux and MacOS.

This is the initial PR in which GitHhub actions are used.

Changes
=======

Use stdlib copyfile instead of copy to only copy the file content.

Fix the error handling for Windows error messages.

Update all the compat tests to check that the errors have valid errors which can be decoded.

As a drive-by change I have marked the Python Windows Service code as obsolete, as is best to use WinSW for handling the windows services.

reviewers: @dumol 

more of a FYI for the i18n fix.

I will create a separate PR for chevah/server where all the tests will be executed on the French Windows...and will try to update the tests here for a better test coverage.


adiroiban:
just a fyi... the win-i18n slave is great is for compat is best to keep it as a separate slave.

for chevah/server, once it is green we can move it as a normal slave for win-2019...but also keep it as unstable to allow to target it directly, if there are issues 

needs-review

dumol:
> just a fyi... the win-i18n slave is great is for compat is best to keep it as a separate slave.
> 
> for chevah/server, once it is green we can move it as a normal slave for win-2019...but also keep it as unstable to allow to target it directly, if there are issues

Ack. Bon travail! ;-]

Have also added some really minor FreeBSD fixes that I've done in my free time. This would make it possible to enable `compat` tests in `python-package` under FreeBSD too.

**changes-approved**

adiroiban:
chevah/server PR at https://github.com/chevah/server/pull/1680

PR #100, opened 2020-07-13
Fix GitHub actions and threadpool tests.

adiroiban:
Scope
=====

This tries to fix the compat to be alwasy green on buildbot and github actions..

It fix a bug in our testing helpers.

Changes
=======

Fix the executeReactor implemenation to wait not only for queued threads but also for the active ones.

reviewers: @dumol 

FYI . This was an ugly bug and I kept ignoring it.
I hope it will not be fixed and compat will always be green.

I don't know why coverage is  not working.


adiroiban:
@dumol I am merging this as I want to have it in master and continue testing the python-package brink integration.

You can review the GitHub action as I have introduced caching here.

Thanks!

dumol:
Thanks! This long-standing issue has plagued my `python-package` builds too…

In regards to the `actions/cache` GitHub action, would it be possible to also cache the `cache` sub-dir? From what I can tell, you have only enabled caching for the `build-compat` sub-dir.

PR #102, opened 2020-10-14
Assume /Users on macOS when not buildslave user.

dumol:
Scope
=====

`test_getHomeFolder_posix` fails for chevah/pythia on Travis' macOS.

More at https://travis-ci.com/github/chevah/pythia/jobs/398801185

Changes
=======

Always assume `/Users` on macOS when the user is not `buildslave`.

**Drive-by fixes**:
  * Overwrite existing file when unzipping the Python package tarball.
  * Added experimental Travis CI.
  * Updated brink/pavement stuff from server and pythia, without removing any supported OS.
  * Updated README accordingly.

reviewers: @adiroiban 

Please review changes.

### DIFF
--- a/chevah/compat/tests/normal/test_system_users.py
+++ b/chevah/compat/tests/normal/test_system_users.py
@@ -46,8 +46,7 @@ class TestSystemUsers(CompatTestCase):
         # For buildslave, home folder is in srv.
         if mk.username == 'buildslave':
             self.assertEqual(u'/srv/' + mk.username, home_folder)
-        elif mk.username == 'runner' and self.os_name == 'osx':
-            # We are on macOS GitHub actions.
+        elif self.os_name == 'osx':
             self.assertEqual(u'/Users/' + mk.username, home_folder)
         else:
             self.assertEqual(u'/home/' + mk.username, home_folder)
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,24 @@
+#
+# Experimental Travis-CI
+#
+# For now, the build matrix has a single member: linux-arm64
+# Pinned on the oldest available distribution, to get a generic Linux build.
+
+language: c
+
+jobs:
+  include:
+    - os: linux
+      dist: xenial
+      arch: arm64
+    - os: osx
+      osx_image: xcode10.1
+    - os: osx
+      osx_image: xcode11.3
+
+install: skip
+
+script:
+  - ./brink.sh detect_os
+  - ./brink.sh deps
+  - ./brink.sh test_ci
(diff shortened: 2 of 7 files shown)


chevah/compat/tests/normal/test_system_users.py
```diff
@@ -46,8 +46,7 @@ def test_getHomeFolder_posix(self):
         # For buildslave, home folder is in srv.
         if mk.username == 'buildslave':
             self.assertEqual(u'/srv/' + mk.username, home_folder)
-        elif mk.username == 'runner' and self.os_name == 'osx':
-            # We are on macOS GitHub actions.
+        elif self.os_name == 'osx':
```
adiroiban, line 49: I think that here we need a helper like `self.ci_name` or something like that to detect over which CI do you run the tests.

If env GITHUB_ACTIONS="true" then we have GHA.

We should also detect Travis and buildbot

dumol: But why? `/Users` is the usual home on macOS. The exception for `buildslave` is already there…

adiroiban: let me know if you need help... for buildbot, the env marker is   BUILDBOT=true

adiroiban: I am using the github action hack in chevah/server and would help to have a standardinze way to detect the CI system in the tests...

yes... the mk.username == 'buildslave' is an ugly hack.

with a code like `elif self.ci_name == 'gha'` the code could be understood without the extra need for a comment like `# We are on macOS GitHub actions.`

The idea is to have the code speak for itself and organize the code and name the variables and helpers in such a way that you don't need a comment.

makes sense?

dumol: > I am using the github action hack in chevah/server and would help to have a standardinze way to detect the CI system in the tests...

Fine by me, but for this PR my goals were minimal. Minor drive-by fixes like adding Travis CI are fine, but let's not get carried away too much.


> yes... the mk.username == 'buildslave' is an ugly hack.

Ugly, but needed for as long as we use Buildbot. Don't have an issue with that though.


> with a code like `elif self.ci_name == 'gha'` the code could be understood without the extra need for a comment like `# We are on macOS GitHub actions.`

But that comment is gone, I have removed that special case. There was no need for it anyway and it only broke something else, namely macOS builds everywhere else.

dumol:
**needs-review**

adiroiban (approved):
Great work. Thanks for the update.
I have updated the README with the travis badge.

PR #104, opened 2021-03-07
Refactor isUserInGroup to getGroupForUser to also get the name of the matched group.

adiroiban:
Scope
=====

This is just a refactor of the existing isUserInGroup functionality so also return the name of the first matched groupl

It should make it easier to identify the name of the group without any extra system calls.

Changes
=======

Use the same implementation but rename the API to getGroupForUser and return the name of the matched group.

In theory, the API could also return the group ID or UID on windows...but I don't see any use of that, so I will let that change for a later time.

As a drive by I have updated the build briink.sh script.

reviewers: @danuker 

Let me know if the changes make sense.

The API is used here https://github.com/chevah/server/pull/1772

-----

This PR was created without a Trac ticket as this is a public repo and I think that for public repo we should do as much work as possible in the public.


adiroiban:
needs-review

danuker (approved):
I pushed a minor test fix, where an assertEquals replaces a forgotten assertTrue (with a tuple of two items, was always true). Also some wording fixes.

Should we listen to semver and increment the major version number?

danuker:
needs-changes

Sorry for not looking at the CI outcomes.

Coverage here does not seem to work. Should we disable it in this repo?
Also, it seems the linter does not work either, because of the way the repository was checked out. Can we check it out by branch name?

setup.py
```diff
@@ -1,7 +1,7 @@
 from setuptools import Command, find_packages, setup
 import os
 
-VERSION = '0.58.3'
+VERSION = '0.59.0'
```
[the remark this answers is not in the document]
adiroiban: compat is still alpha stage.. so it will stay as major v0.

I am not happy with the API ... the representation of paths as segments is a mess.

But maybe we will do a v1 release once it is ported to py3 ... and update to API to something similar to py3 pathlib  https://docs.python.org/3/library/pathlib.html

adiroiban:
Thanks for the review. I will check the CI. There was no other build for compat since 9 Nov 2020 ...

---------

Coverage diff reports works https://app.codecov.io/gh/chevah/compat/compare/104/diff 

I need to delete the domain code.. and move the manual tests code outside of the automated tests.

--------

Will come back with a better PR soon :)

adiroiban:
Thanks for the review.

I have removed the GitHub static check for now. I Will enable it in a separate PR.

We still have buildbot/compat-os-independent which is the old static check

needs-review

adiroiban:
Looks like I still need to fix some coverage https://app.codecov.io/gh/chevah/compat/compare/104/diff

sorry for the noise

needs-changes

chevah/compat/nt_users.py
```diff
@@ -193,10 +193,13 @@ def userExists(self, username):
...
             self.raiseFailedtoCheckUserExists(username, error_text)
 
-    def isUserInGroups(self, username, groups, token):
+    def getGroupForUser(self, username, groups, token):
         """
-        Return true if `username` is a member of `groups`.
+        See: IOSUsers
         """
+        if not groups:
+            raise ValueError('Groups for validation can\'t be empty.')
```
adiroiban, line 201: This is the new change since previous review.

I don't know if we should just remove this code and have None returned on empty list + and error when groups is None.

adiroiban:
Please give it a quick look

I don't want to spend much time on this PR as we should soon start rewriting a lot of the code from here for py3 porting

needs-review

chevah/compat/nt_users.py
danuker: What I would generally expect would be the set theory interpretation (if you remember my JS test about a set having the empty set as a subset).

But the question is, if there are no allowed groups, should we deny access to everyone? In that case, it seems to me a misconfiguration, and it would be hard to debug. But this special ValueError shows us (and possibly the users if they read logs) where to look.

While this introduces a slight inconsistency according to set theory, it would definitely aid in debugging IF we ever call the method with no groups. So, I think this is practical.

chevah/compat/unix_users.py
```diff
@@ -164,14 +171,17 @@ def isUserInGroups(self, username, groups, token=None):
...
             if user_struct.pw_gid == group_struct.gr_gid:
-                return True
+                # Match on group ID.
+                return group
 
             if username_encode in group_struct.gr_mem:
-                return True
-        return False
+                # Match on group name.
+                return group
```
danuker, line 182: This line, and line 174 above is truly not covered. While I don't eyeball any problems, do you want 100% coverage? Would it be easier to add coverage after porting to Py3?

danuker (approved):
If it's OK to have two lines not covered by tests, then this is fine.

I pushed a typo fix.

adiroiban:
Thanks for the review. It's ok to have missing coverage here... as most probably the whole code will be changed with the py3 migration.

This is one the first python code I wrote...and I focused on getting the Win,Linux.AIX,HPUX low-level functions right...and not focused on getting a clean code.

For example, this method can be very well converted into a method.

I will keep this PR open , in case I found other issues while using the new code in chevah/server

PR #105, opened 2021-03-17
Fix path split for single quotes.

adiroiban:
Scope
=====

This fix a critical bug in which path containing single quotes are not handled corretly.

The single quote is replaced for a path separator.

For example, with the old code a path line 
```
/some/O`Reilly/file

was converted in 

/some/O/Reilly/file
```

Changes
=======

Update the path split code and simplify it with explicit regrex escaping.

I have no idea why I have used `%r` there.

reviewers: @danuker 

For now this is a quick fix as this is urgent

In the future we should start using https://hypothesis.readthedocs.io/en/latest/index.html for unit testing in this low level libraries so that we can generate all possible characters and make sure path splinting works for them.

### DIFF
--- a/chevah/compat/posix_filesystem.py
+++ b/chevah/compat/posix_filesystem.py
@@ -112,11 +112,17 @@ class PosixFilesystemBase(object):
                 )
 
     def _pathSplitRecursive(self, path):
-        '''Recursive split of a path.'''
-        separators = os.path.sep
-        if os.path.altsep:
-            separators += os.path.altsep
-        segments = re.split('[%r]' % separators, path)
+        """
+        Recursive split of a path.
+        """
+        if os.path.sep == '\\':
+            # We are on Windows.
+            # Also handle Unix separators and escape the regex.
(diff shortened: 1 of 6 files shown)


chevah/compat/posix_filesystem.py
```diff
@@ -112,11 +112,17 @@ def _impersonateUser(self):
...
-        if os.path.altsep:
-            separators += os.path.altsep
-        segments = re.split('[%r]' % separators, path)
+        """
+        Recursive split of a path.
+        """
+        if os.path.sep == '\\':
+            # We are on Windows.
+            # Also handle Unix separators and escape the regex.
+            separators = '[\\\\/]'
```
danuker, line 121: Would like `r'[\\/]'` better here. But this doesn't stop the merge.

danuker (approved):
Looks OK to me. Tests passed on Windows, though codecov seems to complain about that line still.

I only had a minor comment about the `r` literal.

PR #110, opened 2021-04-29
Update for Twisted 20.3.0

adiroiban:
Scope
=====

This updates the compat testing helpers to work with latest Twisted that support Python 2.7.

We were using a very old Twisted version which was a mix of 12.1 and 15.1

With this change,  any future upgrade to Twisted should not require and update to compat.

Changes
=======

Update the test case.

Move `callCleanup` code fix from chevah/server 

As a drive by, move the static checks to GitHub Actions.
I have also removed the custom skip-ci code as GHA has native support now.
I have also update the brink deps... just in case.

reviewers: @danuker 

FYI.

This will be used in a PR in chevah/server where  will implement Twisted update.

### DIFF
--- a/chevah/compat/testing/testcase.py
+++ b/chevah/compat/testing/testcase.py
@@ -1016,7 +1016,7 @@ class ChevahTestCase(TwistedTestCase, AssertionMixin):
         for function, args, kwargs in reversed(self.__cleanup__):
             try:
                 function(*args, **kwargs)
-            except Exception as error:
+            except Exception as error:  # noqa:cover
                 self._teardown_errors.append(error)
 
         self.__cleanup__ = []
@@ -1046,7 +1046,7 @@ class ChevahTestCase(TwistedTestCase, AssertionMixin):
             except AssertionError as error:
                 errors.append(error.message)
 
-        if errors:
+        if errors:  # noqa:cover
             self._teardown_errors.append(AssertionError(
                 'There are temporary files or folders left over.\n %s' % (
                     '\n'.join(errors))))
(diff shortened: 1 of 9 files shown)


adiroiban:
I think that this is ready for a first  needs-review

I still need to fix static-checks now that they are moved to GitHub ...and fix AIX 7.1... as we use the new VM

chevah/compat/testing/testcase.py
```diff
@@ -1006,9 +1009,16 @@ def addCleanup(self, function, *args, **kwargs):
...
+
+        If a cleanup fails, the next cleanups will continue to be called and
+        the first failure is raised.
         """
         for function, args, kwargs in reversed(self.__cleanup__):
-            function(*args, **kwargs)
+            try:
+                function(*args, **kwargs)
+            except Exception as error:  # noqa:cover
+                self._teardown_errors.append(error)
```
danuker, line 1019: How about appending a `traceback.format_exc()` instead of the error itself?

danuker (approved):
I approve in general; but I would like the tracebacks of the errors also. Do you think they would be too noisy?

chevah/compat/testing/testcase.py
adiroiban: True. Thanks.

dumol:
As I'm trying to get rid of Alpine Linux VMs, have added a workflow to test it under Docker.

As I have found more Buildbot OS'es were being required from Buildbot (but missing because they were recenlty removed), I've added them back through Docker: Amazon Linux 2, CentOS 7/8.

While at it, I've also added Amazon Linux 2018.03 and CentOS 8.2.

Some tests fail consistently on CentOS 8.x (and they are not the same on 8,2 and latest 8).

adiroiban:
Thanks @dumol for the update. This is a great start.

Maybe some tests are failing due to umask differences.

But I think that the remaining can be fixed as part of the py3 migration.

PR #112, opened 2021-06-28
Add support for stacked cleanups.

adiroiban:
Scope
=====

This add support for calling cleanup based on stacks/layers.

This is used by our chevah/server tests in which we have a shared service layer that is used by multiple tests.

Changes
=======

Add a simple enterCleanup/exitCleanup APi for cleanup layers.

As a drive-by change I have updated the IPv4 and IPv6 mocks to return an IP address and not a hostname.

reviewers: @danuker 

check that changes make sense and they can be used in chevah/server

### DIFF
--- a/chevah/compat/testing/testcase.py
+++ b/chevah/compat/testing/testcase.py
@@ -963,6 +963,7 @@ class ChevahTestCase(TwistedTestCase, AssertionMixin):
     def setUp(self):
         super(ChevahTestCase, self).setUp()
         self.__cleanup__ = []
+        self._cleanup_stack = []
         self._teardown_errors = []
         self.test_segments = None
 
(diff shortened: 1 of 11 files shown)


chevah/compat/testing/testcase.py
```diff
@@ -1020,6 +1021,20 @@ def callCleanup(self):
...
+    def enterCleanup(self):
+        """
+        Called when start using stacked cleanups.
+        """
+        self._cleanup_stack.append(self.__cleanup__)
+        self.__cleanup__ = []
+
+    def exitCleanup(self):
+        """
+        To be called at the end of a stacked cleanup.
```
danuker, line 1033: This works well, but could we forget to call `exitCleanup`? 

If you feel like it, perhaps add an assertion somewhere (but not sure where). Or perhaps a context manager.

danuker (approved):
The code looks good (but I have found one possible improvement, but not sure if worth it).

I have tested the cleanup stack with `test_operations` in `server`, and it works. Thanks!

adiroiban:
Thanks for the review. 
I think that we can add a context manager, but these are low-level method and would like to keep them.
But yes, a context manager could make it easier to use.

I would loved to add auto-release on tag...but this is a public repo without access to our VPN.

I will update the context manager and will release it.

danuker:
Works and looks good! Thank you. Will release as `0.61.0`.

adiroiban:
Released
Submitting dist/chevah_compat-0.61.0-py2-none-any.whl to http://deag.chevah.com:10042
Server response (200): OK

PR #114, opened 2021-07-21
[#5687] assertEqual checks Unicode vs. strings in collections

danuker:
Scope
=====

`ChevahTestCase.assertEqual` now goes inside dicts, lists, tuples, and sets to differentiate Unicode vs. raw string objects.

Changes
=======

Added behavior to `chevah/compat/testing/testcase.py`.

Overrode `unittest` methods named `assertSequenceEqual`, `assertSetEqual`, and `assertDictEqual`.

As a drive-by found a bug where `temp_segments` would return raw strings and not Unicode ones.

reviewers: @adiroiban 

Use assertEquals in various ways.

This PR should have a counterpart in `chevah/server` before merging, which fixes the tests broken (or bugs discovered) by the `compat` changes.

### DIFF
--- a/chevah/compat/nt_filesystem.py
+++ b/chevah/compat/nt_filesystem.py
@@ -130,7 +130,7 @@ class NTFilesystem(PosixFilesystemBase):
         else:
             # Default tempfile.gettempdir() return path with short names,
             # due to win32api.GetTempPath().
-            return self._pathSplitRecursive(
+            return self.getSegmentsFromRealPath(
                 win32api.GetLongPathName(win32api.GetTempPath()))
 
     @property
@@ -261,6 +261,8 @@ class NTFilesystem(PosixFilesystemBase):
 
         if path is None or path == u'':
             return segments
+
+        path = text_type(path)
 
         target = os.path.abspath(path.replace('/', '\\')).lower()
         for virtual_segments, real_path in self._avatar.virtual_folders:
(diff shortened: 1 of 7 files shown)


danuker:
I think this is ready for review. Should I open another ticket for the changes on `chevah/server`?

needs-review

danuker:
The windows tests failed. Taking it back; sorry for the noise. needs-changes

chevah/compat/nt_filesystem.py
```diff
@@ -262,6 +262,8 @@ def getSegmentsFromRealPath(self, path):
         if path is None or path == u'':
             return segments
 
+        path = text_type(path)
```
danuker, line 265: This ensures the resulting segments are Unicode and not bytes, using `six`.

danuker:
Done fixing the tests. Changes are ready for review again. needs-review

chevah/compat/nt_filesystem.py
adiroiban: I think that the current code is ok.

adiroiban (approved):
All good. Thanks. Let me know if you need help releasing this.

For chevah/server we need a separate  PR to update chevah/compat.

danuker:
Released as `chevah_compat-0.62.0-py2-none-any.whl`.

PR #116, opened 2021-08-08
Fix docker tests.

adiroiban:
Scope
=====

This is a small PR to fix the docker tests on centos.

Changes
=======

set umask like chevah/server

use /usr/bin/curl for a reference binary.

The bin/cp was a small file on centos and could be confused with a symlink.

I have also updated the compat brink.sh and python version.

----------

On alpine, PAM usage is crashing big times, with segfault.
For now, I have just disable it as I don't plan to use PAM on alpine.

reviewers: @dumol

Scope
=====

This is a small PR to fix the docker tests on centos and alpine.

The centos issue was minor.
The alpine is big as trying to call PAM will segfault. For now we just skip PAM on Alpine.


adiroiban:
@dumol  please take a look. thanks

needs-review

dumol:
Thanks, looks good!

`win-2012` testing still missing, but you can make the repo private and test windows through  our VMs, I guess…

**changes-approved**

adiroiban:
Thanks. I am merging this. 
I think win 2012 should be fine... as is not much different than 2016 .... waiting to see if Win Server 2022 will be a big change.

Issue #680, opened 2022-10-17
Get group membership for Windows users from secondary domain


adiroiban:
The current cpde authenticates users for secondary domain

but then fails to get the controller for the secondary domain in order to get the groups.

PR #681, opened 2022-10-17
[#680] Add support to get groups for users from secondary domain

adiroiban:
Scope
=====

Authentication doesn't fully work for users from a secondary domain.

The credentials are validated, but then getting the groups fails.

Changes
=======

Run the group opration in impersonation mode as the newly authenticated user.

**Drive-by changes**:
  * Updated brink stuff from `python-package` repo, bringing support for more Alpine versions and RHEL 9.
  * Extended the list of tested distros accordingly.

reviewers: @danuker 

FYI

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -83,8 +83,8 @@ jobs:
         sed -i s@^#baseurl=http://mirror.centos.org/\$contentdir/\$releasever/@baseurl=https://vault.centos.org/8.2.2004/@ /etc/yum.repos.d/*.repo
         yum -y upgrade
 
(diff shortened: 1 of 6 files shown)


adiroiban:
@dumol  not urgent. When you have time, can you check why the docker tests are failing


looks like this fails due to our github checkout process

dumol:
> looks like this fails due to our github checkout process

Known issue, documented at https://github.com/chevah/compat/blob/master/.github/workflows/docker.yml#L105.

adiroiban:
This was deployed to customer, but the issue was not fixed.

PR #683, opened 2022-12-05
[#682] Don't lock the file when open for reading.

### LINKED ISSUE #682: Don't lock the file on read on Windows
The compat should try to provide the same behaviour over linux and windows.

Under linux, when a file is opened, it is not locked, so it can be removed or writte.

On Window, there is a lock, as this is the default Python behaviour.

adiroiban:
Scope
=====

Fixes #682

This tries to get the same behaviour over linux and Windows when opening a file for reading.

The default Python code locks the file on Windows, but the file is not locked on Linux.

Changes
=======

Implement custom Windows open code.

reviewers: @danuker 

Check that changes make sense

I will create a separate PR for chevah/server in which this will be used and we can do more testing.

### DIFF
--- a/chevah/compat/nt_filesystem.py
+++ b/chevah/compat/nt_filesystem.py
@@ -11,6 +11,7 @@ from six.moves import range
 from contextlib import contextmanager
 from winioctlcon import FSCTL_GET_REPARSE_POINT
 import errno
+import msvcrt
 import ntsecuritycon
 import os
 import pywintypes
@@ -978,3 +979,42 @@ class NTFilesystem(PosixFilesystemBase):
                 if group_sid == sid:
                     return True
         return False
+
+
+    def openFile(self, segments, flags, mode):
+        """
+        See `ILocalFilesystem`.
+        """
+        path = self.getRealPathFromSegments(segments, include_virtual=False)
+        path_encoded = self.getEncodedPath(path)
+
+        self._requireFile(segments)
+        with self._convertToOSError(path), self._impersonateUser():
+            # FIXME: todo
+            return os.open(path_encoded, flags, mode)
+
+    def openFileForReading(self, segments):
+        """
+        See `ILocalFilesystem`.
+        """
+        path = self.getRealPathFromSegments(segments, include_virtual=False)
+        path_encoded = self.getEncodedPath(path)
+
+        self._requireFile(segments)
+        with self._convertToOSError(path), self._impersonateUser():
+
+            handle = win32file.CreateFile(
+                path_encoded,
+                win32file.GENERIC_READ,
+                win32file.FILE_SHARE_DELETE | win32file.FILE_SHARE_READ,
+                None,
+                win32file.OPEN_EXISTING,
+                0,
+                None,
+                )
+
+            # detach the handle
+            detached_handle = handle.Detach()
+            fd = msvcrt.open_osfhandle(
+                detached_handle, os.O_RDONLY)
+            return os.fdopen(fd, 'rb')
--- a/chevah/compat/tests/normal/test_filesystem.py
+++ b/chevah/compat/tests/normal/test_filesystem.py
@@ -1715,6 +1715,25 @@ class TestLocalFilesystem(DefaultFilesystemTestCase):
             if a_file:
                 a_file.close()
 
+    def test_openFileForReading_no_delete_lock(self):
+        """
+        A file opened only for reading will not be locked for delete
+        operations.
+        """
+        _, segments = mk.fs.makePathInTemp()
+        mk.fs.writeFileContent(segments=segments, content='something-\N{sun}')
+        a_file = None
+        try:
+            a_file = self.filesystem.openFileForReading(segments)
+
+            self.filesystem.deleteFile(segments)
+
+            content = a_file.read(100)
+            self.assertEqual(b'something-\xe2\x98\x89', content)
+        finally:
+            if a_file:
+                a_file.close()
+
     def test_openFileForWriting_ascii(self):
         """
         Check opening a file for writing in plain/ascii/str mode.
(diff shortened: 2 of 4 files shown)


adiroiban:
This is ready for a review.

No need to run manual tests.

We have the automated tests that are also executed on Windows

needs-review

chevah/compat/tests/normal/test_filesystem.py
```diff
@@ -1715,6 +1715,95 @@ def test_openFileForReading_no_write(self):
...
+        mk.fs.writeFileContent(segments=segments, content=content)
+        fd = None
+        try:
+            fd = self.filesystem.openFile(
+                segments,
+                flags=self.filesystem.OPEN_READ_ONLY,
+                mode=0,
+                )
+
+            self.filesystem.deleteFile(segments)
```
danuker, line 1755: Maybe also check that the file WAS deleted (such as with `os.stat`).

chevah/compat/tests/normal/test_filesystem.py
```diff
@@ -1715,6 +1715,95 @@ def test_openFileForReading_no_write(self):
...
+            result = os.read(fd, 100)
+            self.assertEqual(content.encode('utf-8'), result)
+        finally:
+            if fd:
+                os.close(fd)
+
+    @conditionals.onOSFamily('nt')
+    def test_fdRead_os_error(self):
+        """
+        This is a low-level Windows tests.
```
danuker, line 1766:
```suggestion
        This is a low-level Windows test.
```

chevah/compat/tests/normal/test_filesystem.py
```diff
@@ -1715,6 +1715,95 @@ def test_openFileForReading_no_write(self):
...
+
+        Under normal update, this function should never fail as we have some
+        higher API guards.
+
+        This tests is here to make sure that if it fails, it raises a generic
```
danuker, line 1771:
```suggestion
        This test is here to make sure that if it fails, it raises a generic
```

chevah/compat/tests/normal/test_filesystem.py
```diff
@@ -1715,6 +1715,95 @@ def test_openFileForReading_no_write(self):
...
+
+            # File can't be read.
+            self.assertRaises(OSError, os.read, fd, 1)
+
+        finally:
+            if fd:
+                os.close(fd)
+
+        # Just check the result after close.
+        self.assertEqual(content, mk.fs.getFileContent(segments))
```
danuker, line 1805: I'd like an `OPEN_READ_WRITE` test also, as I said in `chevah/compat/nt_filesystem.py`.

danuker (changes requested):
I'm a bit worried about the `OPEN_READ_WRITE` case.

chevah/compat/nt_filesystem.py
```diff
@@ -978,3 +979,64 @@ def hasGroup(self, segments, group):
...
+        See `ILocalFilesystem`.
+
+        `mode` is ignored on Windows.
+        """
+        path = self.getRealPathFromSegments(segments, include_virtual=False)
+        path_encoded = self.getEncodedPath(path)
+
+        self._requireFile(segments)
+        with self._convertToOSError(path), self._impersonateUser():
+            if (
```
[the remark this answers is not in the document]
adiroiban: Yes. On Windows, os.open() from STDLib does an exclusive lock and there is no flag to disable this behaviour :(

chevah/compat/tests/normal/test_filesystem.py
adiroiban: I think this was ok.
If `filesystem.deleteFile` is not doing what it should, we should have a separate test and fix for the `delteFile` method.

To be extra safe, I went with:

```diff
            self.filesystem.deleteFile(segments)
+           self.assertFalse(self.filesystem.exists(segments))
```

chevah/compat/tests/normal/test_filesystem.py
adiroiban: Added.  Thanks! Good catch!

chevah/compat/nt_filesystem.py
```diff
@@ -978,3 +979,64 @@ def hasGroup(self, segments, group):
...
+                flags & self.OPEN_READ_ONLY == self.OPEN_READ_ONLY
+                and flags & self.OPEN_WRITE_ONLY != self.OPEN_WRITE_ONLY
```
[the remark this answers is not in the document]
adiroiban: Yes. Good catch. I patched it as best as I could :(

adiroiban:
Many thanks for another great review.

I don't know why coverage is not happy, but we can deal with it later.

I hope I have address all the major issues

needs-review

danuker (approved):
Looks good! Thanks.
I pushed a typo fix.

chevah/compat/nt_filesystem.py
```diff
@@ -978,3 +999,69 @@ def hasGroup(self, segments, group):
...
+    def _fdRead(self, path):
+        """
+        Do the low-level Windows file open.
+
+        Returns a file descriptor.
+        """
+        desired_access = win32file.GENERIC_READ
+        share_mode = (
+            win32file.FILE_SHARE_DELETE |
+            win32file.FILE_SHARE_WRITE |
```
adiroiban, line 1046: @danuker in previous version I was missing FILE_SHARE_WRITE ... so it was failing to open a file that was already opened for writing.

This should get a behaviour closer to what python os.open does.

I tried to do a quick search for python implementation of `open` but I could not find it

adiroiban:
Thanks for the review.

There were a few changes and updated tests.

The failing tests were observed on chevah/server

I have implemented more tests into chevah/compat. I hope this should give us a better coverage for the intended behaviour

needs-review

danuker (approved):
Looks good. I pushed a few typo fixes.

adiroiban:
Thanks. I have merged this without a new-release.

The 0.65.5 is already used in chevah/server for a long time

PR #693, opened 2023-05-24
[#692] Add support for Python 3.11.

adiroiban:
Scope
=====

Fixes #692

This updates compat to work with Python 3.11.

Changes
=======

Use a newer brink that contains the nose patching for 3.11

How to try and test the changes
===============================

reviewers: @dumol  

more of a FYI

this is to make sure CI is green on 3.11

PR #700, opened 2024-10-04
[#699] Python 3.12 update.

### LINKED ISSUE #699: Python 3.12 update.
Try to see what it would take, esp. regarding the situation with `nose` and `imp`…

dumol:
Scope
=====

Fixes #699 

Updates for Python 3.12.

Changes
=======

Use `pynose` and `importlib.utils` instead of `nose` and `imp`.

**Drive-by changes**:
  * Updated `pythia.*` to latest versions from pythia repo.
  * Simplified cloning independently under Docker, now that CentOS 5 is no longer relevant.

reviewers: @adiroiban 

Have only tried to have things green and to follow #693.

Do check if the diff looks alright.


pavement.py
```diff
@@ -120,7 +120,7 @@ def compile_file(fullname, ddir=None, force=0, rx=None, quiet=0):
             if not force:
                 try:
                     mtime = int(os.stat(fullname).st_mtime)
-                    expect = struct.pack('<4sl', imp.get_magic(), mtime)
+                    expect = struct.pack('<4sl', MAGIC_NUMBER, mtime)
```
adiroiban, line 123: thanks for the update... the compile_file code shoudl be removed ... and just use upstream code


We had this "fork" here to support py2 migration... and I forgot to clean the code

adiroiban (approved):
Changes looks good. Great work with py3.12 and compat updates.

I did more py3 only cleanup, while doing the review.

I also updated the GHA YAML to run the tests on latest Ubuntu

I tried to simplify the docker YAML code as without older centos, we should be able to use standard GitHub Actions rules.

I have also fixed a bug in the testing code, so we can have a new release

adiroiban:
I see that caching failes to save for alpine ... due to busybox tar file

https://github.com/chevah/compat/actions/runs/11202300894/job/31138241218?pr=700#step:24:4


Can we get GNU tar installed on alpine ?

.github/workflows/ci.yaml
```diff
@@ -265,7 +269,7 @@ jobs:
 
     - name: Alpine setup
       if: startsWith(matrix.container, 'alpine')
-      run: apk add git bash shadow sudo curl
+      run: apk add git bash shadow sudo curl tar
```
adiroiban, line 272: Cache save was failing on Alpine due to busybox tar

https://github.com/chevah/compat/actions/runs/11202300894/job/31138241218?pr=700#step:24:4

.github/workflows/ci.yaml
```diff
@@ -288,52 +292,30 @@ jobs:
...
-        # Update repo token.
-        git remote add origin https://github.com/chevah/$CHEVAH_REPO
-        git fetch --no-tags --prune origin
-        # Prepare the code.
-        git clean -f
-        git reset --hard ${{ github.event.after }}
-        git log -1 --format='%H'
-
-    # This action also fails on CentOS 5/6.
+    - uses: actions/checkout@v4
```
adiroiban, line 295: I went with default git chechout code ... to simplify our GHA yaml

.github/workflows/ci.yaml
dumol: ACK

.github/workflows/ci.yaml
dumol: The downside is it only supports distributions on which GitHub Actions Runner actually runs.

dumol:
There's one more thing needed before a new `compat` release: production packages of Pythia. This PR has been using testing packages generated in https://github.com/chevah/pythia/pull/58, which is not yet merged.

.github/workflows/ci.yaml
adiroiban: Well... it works on any distribution on which nodejs 16 or 20 can be executed.

I have changed the code and it looks like it also works for Alpine ...for example

GitHub Actions has a separate node_js version compiled for Alpine.

GitHub Actions are doing with nodejs, something similar to what we do with  pythia... they distribute various nodejs versions with each GitHub Action runner.

------

you can mark this as resolved and merge

adiroiban:
Yes... for chevah/compat is not that critical..

I am a bit worried about the new libnsl dependency ... but I see that the code works on OracleLinux 8 

-------

I think that we should update the GitHub Actions on Pythia to run some minimal tests via Docker on multiple centos versions ... just to make sure we got our dynamic libraries right

dumol:
> I think that we should update the GitHub Actions on Pythia to run some minimal tests via Docker on multiple centos versions ... just to make sure we got our dynamic libraries right

Tracked at https://github.com/chevah/pythia/issues/68.

PR #704, opened 2024-11-03
[#689]  Fix temp folder cleanup. Enable dependabot

### LINKED ISSUE #689: Setup dependebot
Something like this

https://github.com/twisted/klein/pull/597

I will also enable more security checks for GitHub

We can start with github actions, and then we can look for pypi updates (After the py3 migration)

adiroiban:
Scope
=====

Fixes #689

The temp folder on macOS is not cleaned with Python 3.12

Also enable dependabot to keep deps up to date.

Changes
=======

Allow cleanFolder in temp dir

Enable dependbot montly updates.

As a drive by, migrated from setup.cfg to pyproject.toml. dependbot should pick up the deps from there. No need for separate requirement.txt file

reviewers: @dumol 

More of a fyi


adiroiban:
needs-review

dumol (approved):
> More of a fyi

ACK

PR #710, opened 2024-12-08
[#709] Add support for pytest 

### LINKED ISSUE #709: Make code compatible with pytest

adiroiban:
Scope
=====

Fixes #709

This add support to run compat tests with `pytest`

`pytest` is the current de-facto test framework for Python and we should start using it as it has many nice features. `nose` and `pynose` are dying.

Changes
=======

Update tests for py3

Add pytest configuration.

Run all the `chevah-compat` tests also with `pytest` not only with `pynose`.

As a drive by, update more code for py3 only.

reviewers: @dumol 

More of a fyi

1.Normal tests
-----------------

You can now run compat test using `pytest` as either

* `./pythia.sh pytest`
* or just run `pytest` directly via `./build-py3/bin/pytest`

2.Elevated tests
-----------------

There are also the "elevated" tests for OS users that are triggered via `sudo ./pythia.sh pytest src/chevah_compat/tests/elevated/`

during test run they will automatically create OS users and will delete the users and group at the end of the test

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -241,6 +241,20 @@ jobs:
         CODECOV_TOKEN: enabled
         USER: runneradmin
 
+    - name: Test pytest
+      run: |
+        ./pythia.sh pytest
+      env:
+        CODECOV_TOKEN: enabled
+        USER: runneradmin
+
+    - name: Test pytest elevated
+      run: |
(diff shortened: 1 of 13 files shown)


adiroiban:
needs-review

dumol (approved):
Fine by me!

I've noticed the new `pytest` tests are only running on the old Ubuntu and Windows. Is this intended?

PR #719, opened 2025-03-19
#718 Allow PAM auth as non root user

### LINKED ISSUE #718: PAM auth without root

adiroiban:
Scope
=====

Fixes #718

This adds support to authenticate Linux OS users using only PAM as non root.

This is somehow similar to what we get in WIndows.

I am user `chevah` , as long as I get username and password for user `JohnDoe` I can use PAM, as a non root user, to check whether the credentials are valid.

Changes
=======

Add a separeate non-root PAM only method.

The old root PAM method was kept for backward compatibility

Automated tests are complicated... when we autoamtically create OS test users we run as root.

This functionality assumes that you are not root... but if we are not root, we can't automatically add OS users... so we can't do automated tests

I wrote some tests to make sure we don't get unexpected errors for wrong password

--------

onOSVersion was removed as the low-level API no longer support is.
os_version was moved to be available only during testing.
We might add it back later.

----------

As a drive-by change I have updated some deps

reviewers: @dumol 

More of a FYI

There will be some manual tests in chevah/server

This is also hard to test.

Most PAM setup, will only allow validating the password of the current user... for example to help with screensavers.
If you need to authenticate other users, you will need root access.

THis is unfortunate... but this is Linux.

Some PAM setup, might have a LDAP or SSSD module that does the authentication over the network and root is not required.

The code from here is to provide the low-level API.

We will see in chevah/server how to implement this in an application

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -14,7 +14,7 @@ coverage:
     project:
       default:
         target: auto
-        threshold: 0.05%
+        threshold: 0.1%
 
 comment:
   layout: "header, diff, files, components"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "chevah-compat"
-version = "1.4.0"
+version = "1.5.0"
 maintainers = [
     { name = "Adi Roiban", email = "adi.roiban@proatria.com" },
 ]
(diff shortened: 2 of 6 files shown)


adiroiban:
main PR at https://github.com/chevah/server/pull/7042

But that PR is not really using this code yet.

It looks like Linux OS is not yet ready to run in non-root mode... something that Windows already does for a very long time

needs-review

dumol (approved):
> More of a FYI

ACK

adiroiban:
Thanks for checking this.
I am merging this as 1.5.0 was released

